PR #1, opened 2023-04-06
Implement basic flask application management layer

weiiwang01:
Add the capability of starting a simple sample flask application.


sample_rock/rockcraft.yaml
```diff
@@ -0,0 +1,37 @@
...
+      chmod 777 $CRAFT_OVERLAY/tmp
+      groupadd -R $CRAFT_OVERLAY --gid 2000 flask
+      useradd -R $CRAFT_OVERLAY --system --gid 2000 --uid 2000 --home /srv/flask flask
+  flask-dependencies:
+    stage-packages:
+      - python3-dev
+    build-packages:
+      - build-essential
+      - python3-pip
+    plugin: nil
```
arturo-seijas, line 29: I'm wondering if the `python` plugin would work here

weiiwang01: I tried to use the python plugin, but the python plugin does not work properly without a `setup.py`.

arturo-seijas (dismissed):
LGTM

sample_rock/rockcraft.yaml
```diff
@@ -0,0 +1,38 @@
...
+    overlay-script: |
+      mkdir -p $CRAFT_OVERLAY/srv/flask/.local/lib/python3.10/site-packages
+      pip install flask gunicorn --target $CRAFT_OVERLAY/srv/flask/.local/lib/python3.10/site-packages
```
gregory-schiano: As discussed I'm not in favor of installing flask and gunicorn using pip as it's out of Canonical's control
Installing it through apt will perhaps give older versions, but version that we can easily manage and support for long term.

So unless we decide to host our own pip repo, I'd install these deps using apt.
Note that if the user want to override this and install through pip they will be able to do it, so it won't block the user.

src/charm.py
```diff
@@ -5,16 +5,60 @@
...
+            The pebble layer definition for flask application.
+        """
+        return {
+            "services": {
+                "flask-app": {
+                    "override": "replace",
+                    "summary": "Flask application service",
+                    "command": "python3 -m gunicorn --chdir /srv/flask/app app:app -b 0.0.0.0:8080",
+                    "user": "flask",
+                    "group": "flask",
```
gregory-schiano, line 56: Will gunicorn worker process will also be started using this user/group ? Shouldn't we also specify `user` and `group` to gunicorn ?

src/charm.py
```diff
@@ -5,16 +5,60 @@
...
+
+        Returns:
+            The pebble layer definition for flask application.
+        """
+        return {
+            "services": {
+                "flask-app": {
+                    "override": "replace",
+                    "summary": "Flask application service",
+                    "command": "python3 -m gunicorn --chdir /srv/flask/app app:app -b 0.0.0.0:8080",
```
gregory-schiano: Since the tech spec will expect a configuration file, I suggest to specify `chdir`
I would also keep the default port and binding address unless it causes issues for the charm ?

src/charm.py
weiiwang01: I don't think normal user has the capability to create a new process with a different user/group.

src/charm.py
weiiwang01: Yes, I will move these command line arguments into a configuration file in the next story.
The default binding address is localhost, which is not accessible outside the container.

gregory-schiano: Ok can we keep the default port which is 8000 ?

sample_rock/rockcraft.yaml
weiiwang01: Yes, you are right, I will update the code use `python3-flask` and `python3-gunicorn` from the apt source.
The code here can be reused if we want to support installing dependencies from the `requirements.txt` provided from the flask application in the future.

src/charm.py
weiiwang01: Okay, no problem

src/charm.py
```diff
@@ -5,16 +5,60 @@
...
+
+        Returns:
+            The pebble layer definition for flask application.
+        """
+        return {
+            "services": {
+                "flask-app": {
+                    "override": "replace",
+                    "summary": "Flask application service",
+                    "command": "python3 -m gunicorn --chdir /srv/flask/app app:app -b 0.0.0.0:8000",
```
amandahla, line 54: How can we check the application logs? Is it possible to see in the pod? Would be worth changing the log level here?

.trivyignore
```diff
@@ -0,0 +1,9 @@
+# ignore CVEs in python3-flask and python3-gunicorn
```
jdkandersson: Did we discuss that we wouldn't use `python3-flask` which means that at least some of these CVEs are no llonger applicable?

sample_rock/rockcraft.yaml
```diff
@@ -0,0 +1,33 @@
...
+      mkdir $CRAFT_OVERLAY/etc
+      mkdir $CRAFT_OVERLAY/tmp
+      chmod 755 $CRAFT_OVERLAY/etc
+      chmod 777 $CRAFT_OVERLAY/tmp
+      groupadd -R $CRAFT_OVERLAY --gid 2000 flask
+      useradd -R $CRAFT_OVERLAY --system --gid 2000 --uid 2000 --home /srv/flask flask
+  flask-dependencies:
+    stage-packages:
+      - python3-dev
+      - python3-flask
```
jdkandersson: Did we want to change this to use a requirements.txt file?

sample_rock/sample-app/app.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+from flask import Flask
+
+app = Flask(__name__)
+
+@app.route("/")
+def hello_world():
+    return "Hello, World!"
```
jdkandersson, line 10: The flask sample app in the spec was changed to include a `logging.info(json.dumps(os.environ))` to make it easier for the user to know all the available environment variables. Should we make the change here as well? Perhaps also useful for testing?

src/charm.py
```diff
@@ -5,16 +5,60 @@
...
+        self.framework.observe(self.on.config_changed, self.config_service)
+
+    def config_service(self, event: ConfigChangedEvent) -> None:
+        """Configure the flask pebble service layer.
+
+        Args:
+            event: the config-changed event that trigger this callback function.
+        """
+        container = self.unit.get_container("flask-app")
+        if container.can_connect():
```
jdkandersson: to reduce the `if/else`, how about:

```python
if not container.can_connect():
    event.defer()
    return

... # code for being able to connect
```
That means there is less context required for understanding the conditionals

tests/integration/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+    async def _get_unit_ips(application_name: str):
+        """Retrieve unit ip addresses of a certain application.
+
+        Returns:
+            a list containing unit ip addresses.
+        """
+        _, status, _ = await ops_test.juju("status", "--format", "json")
+        status = json.loads(status)
+        units = status["applications"][application_name]["units"]
+        ip_list = []
```
jdkandersson: What about:

```python
ip_list = tuple(units[key]["address"] for key in sorted(units.keys(), key=lambda n: int(n.split("/")[-1])))
```

tests/integration/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+        for key in sorted(units.keys(), key=lambda n: int(n.split("/")[-1])):
```
jdkandersson: Out of curiosity, what is the reason for sorting here?

tests/integration/conftest.py
jdkandersson: using tuple since this should be a constant list

tests/integration/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+"""Fixtures for flask charm integration tests."""
+
+import json
+
+import pytest_asyncio
+from pytest_operator.plugin import OpsTest
+
+
+@pytest_asyncio.fixture(scope="module", name="get_unit_ips")
+async def fixture_get_unit_ips(ops_test: OpsTest):
```
yanksyoon, line 13: Out of curiosity, what is the benefit of returning a function to retrieve unit ip addresses rather over returning the unit IP addresses directly?

tests/integration/test_charm.py
```diff
@@ -8,29 +8,37 @@
...
 
 import pytest
-from pytest import Config
+import pytest_asyncio
+import requests
 from pytest_operator.plugin import OpsTest
 
 logger = logging.getLogger(__name__)
 
+APP_NAME = "flask-k8s"
```
yanksyoon: I'm wondering if `APP_NAME` is required, would it be better if we deploy the application `application = model.deploy(...)` and then use the name from the application directly: `application.name`.

yanksyoon (dismissed):
LGTM, some questions from me, thank you!

src/charm.py
weiiwang01: I think the application log will go to the pod log right now. But I think there will be a story to add COS integration in the future. At that time we can change the log destination.

tests/integration/conftest.py
weiiwang01: This is a fixture, so if the fixture returns the unit ip addresses, it will return the unit ip addresses before the test case executing. Return a function gives us the flexibility of when to retrieve the unit ip addresses, which can be handy if we would like to modify the unit status in the middle of a test cases.

tests/integration/test_charm.py
weiiwang01: Good idea, updated, thanks!

.trivyignore
weiiwang01: Updated, I switched the flask installation to use pip instead of apt.

sample_rock/rockcraft.yaml
weiiwang01: Updated, thanks!

sample_rock/sample-app/app.py
weiiwang01: There will be another endpoint for that in the flask server. I will add that in the future.

tests/integration/conftest.py
weiiwang01: It's sorted by the unit number, so every invocation will has the exactly same result if units are not changed.

tests/integration/conftest.py
weiiwang01: I think it's okay to use a list here, since the list as a return value is not shared between different places.

src/charm.py
weiiwang01: Yes, that's better, updated, thanks!

sample_rock/rockcraft.yaml
gregory-schiano: Well this is debatable. It's not what we want, but what is the standard ?
Do Flask developer expect the flask dependency to be available on the system and therefore keeps it out of their requirements.txt or do they put flask as a dependency.
If they add it, that's fine to expect it in requirements.txt and therefore let the pip install do the job... but if not then we'll have issues.

I suggest to keep it as it is now, but discuss / investigate what's the standard for Flask developers

sample_rock/sample-app/app.py
gregory-schiano: That's ok for me, the sample app is just here to illustrate the spec, the sample app used for our integration tests doesn't need to be exactly the same

arturo-seijas (dismissed):
LGTM

sample_rock/rockcraft.yaml
jdkandersson: I should have added context, the `python3-flask` apt dependency is quite out of date and has security issues. We should add another ticket to our backlog to ask for or fix the situation ourselves, in the meantime we just install using pip. @weiiwang01 could you please create that ticket?

tests/integration/conftest.py
jdkandersson: let's discuss in the standup

weiiwang01: I changed the `get_unit_ips` to return a tuple, thanks!

sample_rock/rockcraft.yaml
weiiwang01: Yes, I will create a followup to investigate this issue.

jdkandersson (approved):

merkata (approved):
Looks good!

PR #2, opened 2023-04-13
Add gunicorn configurations to the flask charm

weiiwang01:
This pull request adds Gunicorn related charm configuration (`webserver_workers`, `webserver_threads`, `webserver_keepalive`, `webserver_timeout`) to the Flask charm.


config.yaml
```diff
@@ -1 +1,15 @@
-options: {}
+options:
+  webserver_workers:
+    type: int
+    description: The number of gunicorn worker processes for handling requests.
```
jdkandersson: These configuration options were intended to be a little more generic, do you think it would be better not to mention gunicorn here?

jdkandersson: Similar comment for the other configuration options

src/charm.py
```diff
@@ -26,20 +28,89 @@ def __init__(self, *args: typing.Any) -> None:
         super().__init__(*args)
         self.framework.observe(self.on.config_changed, self.config_service)
 
+    def flask_container(self, require_connected: bool = True) -> Container:
+        """Get the flask application workload container controller.
+
+        Args:
+            require_connected: if set to ``True``, a runtime exception will be raised if the
```
jdkandersson: single quote?

src/charm.py
```diff
@@ -26,20 +28,89 @@ def __init__(self, *args: typing.Any) -> None:
...
 
         Args:
             event: the config-changed event that trigger this callback function.
         """
-        container = self.unit.get_container("flask-app")
+        container = self.flask_container(require_connected=False)
         if not container.can_connect():
             event.defer()
             return
+        service_name = "flask-app"
```
jdkandersson: Should we add a few empty lines to group similar lines of code?

src/charm.py
```diff
@@ -26,20 +28,89 @@ def __init__(self, *args: typing.Any) -> None:
...
+        gunicorn_config_path = "/srv/flask/gunicorn.conf.py"
```
jdkandersson: this looks like a good candidate for a separate module? Any thoughts? @gregory-schiano as well

src/charm.py
```diff
@@ -26,20 +28,89 @@ def __init__(self, *args: typing.Any) -> None:
...
+        """Write the given text content to a file inside the workload container.
+
+        If the file doesn't exist, a new file will be created. All pre-existent content will be
+        overwritten by this operation.
+
+        Args:
+            path: Path to the file in the workload container.
+            content: the text content to be written to the file.
+        """
+        self.flask_container().push(path, content, encoding="utf-8")
```
jdkandersson: Didn't we intend to use a shared volume for this?

src/charm.py
```diff
@@ -26,20 +28,89 @@ def __init__(self, *args: typing.Any) -> None:
...
+        except PathError:
+            return None
+
+    def gunicorn_config(self) -> str:
+        """Generate the content of the gunicorn configuration file based on charm configurations.
+
+        Returns:
+            The content of the gunicorn configuration file.
+        """
+        config_entries = ["bind = ['0.0.0.0:8000']", "chdir = '/srv/flask/app'"]
```
jdkandersson: Should we manage this using a jinja template potentially in another module? This is also a candidate where we can start to use the charm state abstraction

tests/integration/conftest.py
```diff
@@ -27,4 +51,21 @@ async def _get_unit_ips(application_name: str):
             for _, unit_status in sorted(units.items(), key=lambda kv: int(kv[0].split("/")[-1]))
         )
 
-    yield _get_unit_ips
+    return get_unit_ips
+
+
+@pytest_asyncio.fixture
+async def update_config(model: Model, request: FixtureRequest, flask_app: Application):
```
jdkandersson, line 94: What is this fixture intended to do?

tests/integration/test_charm.py
```diff
@@ -41,3 +26,19 @@ async def test_flask_is_up(flask_app, get_unit_ips):
         response = requests.get(f"http://{unit_ip}:8000", timeout=5)
         assert response.status_code == 200
         assert "Hello, World!" in response.text
+
+
+@pytest.mark.parametrize("update_config", [{"webserver_timeout": 7}], indirect=True)
+@pytest.mark.usefixtures("update_config")
+async def test_flask_webserver_config(
```
jdkandersson: Can we also come up with tests for the other configuration options?

tests/unit/conftest.py
```diff
@@ -3,19 +3,35 @@
...
 @pytest.fixture(name="harness")
-def harness_fixture():
+def harness_fixture(monkeypatch) -> typing.Generator[ops.testing.Harness, None, None]:
     """Ops testing framework harness fixture."""
-    ops.testing.SIMULATE_CAN_CONNECT = True
     harness = ops.testing.Harness(FlaskCharm)
+    orig_send_signal = ops.testing._TestingPebbleClient.send_signal
 
-    yield harness
+    def patched_send_signal(self, sig, service_names):
```
jdkandersson: Out of curiosity, which bug?

tests/unit/conftest.py
```diff
@@ -3,19 +3,35 @@
...
+        return orig_send_signal(self, sig, *service_names)
 
+    monkeypatch.setattr(ops.testing._TestingPebbleClient, "send_signal", patched_send_signal)
+    yield harness
     harness.cleanup()
-    ops.testing.SIMULATE_CAN_CONNECT = False
+
+
+@pytest.fixture
+def mock_container_fs(monkeypatch) -> dict[str, str]:
```
jdkandersson: This is the other reason of using a shared volume and moving the configuration handling to a separate module so the patch isn't needed

jdkandersson:
It could be worth writing a design doc for the modules and interfaces that we should have for this feature

src/charm.py
jdkandersson: @gregory-schiano

config.yaml
gregory-schiano: Indeed, no a big deal but will ease moving this later

src/charm.py
gregory-schiano: Yes definitely, we need a state and a service to manage the file generation from the state and managing the generated file

src/charm.py
```diff
@@ -26,20 +28,89 @@ def __init__(self, *args: typing.Any) -> None:
...
             event: the config-changed event that trigger this callback function.
         """
-        container = self.unit.get_container("flask-app")
+        container = self.flask_container(require_connected=False)
         if not container.can_connect():
             event.defer()
             return
+        service_name = "flask-app"
+        gunicorn_config_path = "/srv/flask/gunicorn.conf.py"
+        current_gunicorn_config = self.pull_file(gunicorn_config_path)
```
gregory-schiano: I would no recommend pulling the file, the event should indicate if the change requires a reload
For example any config change that modify a property starting with "webserver" will require a reload, this will be known in the config_changed event
Then the config_changed event will call the service to generate the file passing the state to it. I think this service could be a "gunicorn_service" that will know where to put the config file, how to generate it, and how to reload the service, and how to know if the reloading was successful

src/charm.py
gregory-schiano: As discussed would be better, let's see how the discussion with Juju team goes concerning the "default volume"

gregory-schiano (commented):
I won't look deeper in this PR, in terms of architecture we need to start implementing the charm architecture we defined with David
For this PR I expect:
- A state class that will contain the gunicorn state information coming from config, or being constants for now
- A gunicorn service which will take care of managing the gunicorn service (outside pebble) and its configuration

src/charm.py
weiiwang01: I think we are using the reStructuredText in docstrings? The reStructuredText use double back quotes for code blocks.

src/charm.py
weiiwang01: I refactored the code to conformed with the standard, thank you.

src/charm.py
weiiwang01: I added some comments in the Charm Complexity document.

src/charm.py
weiiwang01: The configuration file does not have enough boilerplate to justify the use of jinja template.

tests/integration/conftest.py
weiiwang01: Update the charm configuration before the test case and reset the charm configuration after the test case finished.

tests/integration/test_charm.py
weiiwang01: There isn't other configuration option right now, all configurations are web server configurations.

tests/unit/conftest.py
weiiwang01: https://github.com/canonical/operator/pull/926

src/charm.py
weiiwang01: I have refactored the charm to conform with the new standard, thanks!

tests/unit/conftest.py
weiiwang01: Please see the discussion about shared volumes.

config.yaml

src/charm.py

sample_rock/sample-app/app.py
```diff
@@ -1,10 +1,19 @@
...
 @app.route("/")
 def hello_world():
     return "Hello, World!"
+
+
+@app.route("/sleep")
+def sleep():
+    duration_seconds = int(request.args.get("duration"))
+    time.sleep(duration_seconds)
+    return ""
```
arturo-seijas, line 19: It this return necessary?

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
         if not container.can_connect():
             event.defer()
             return
+
+        service_name = "flask-app"
         container.add_layer("flask-app", self.flask_layer(), combine=True)
+        webserver_config_path = str(self._webserver.config_path)
+        current_webserver_config = self.pull_file(webserver_config_path)
+        is_webserver_running = container.get_service(service_name).is_running()
+        if current_webserver_config != self._webserver.config:
```
arturo-seijas: IS it worth to check if the configuration has changed? This is only triggered when the config changes, which is not very often, so IMHO the added complexity is not justified

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
+        self._charm_state = CharmState(charm=self)
+        self._webserver = GunicornWebserver(charm_state=self._charm_state)
         self.framework.observe(self.on.config_changed, self.config_service)
 
+    def flask_container(self, require_connected: bool = True) -> Container:
+        """Get the flask application workload container controller.
+
+        Args:
+            require_connected: if set to ``True``, a runtime exception will be raised if the
+                pebble inside the container is not ready.
```
arturo-seijas: s/pebble/pebble service?

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
+
+        Return:
+            The controller of the flask application workload container.
+
+        Raises:
+            RuntimeError: if the pebble inside  the container is not ready while the
```
arturo-seijas: s/pebble/pebble service?

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
 
+    def push_file(self, path: str, content: str) -> None:
+        """Write the given text content to a file inside the workload container.
+
+        If the file doesn't exist, a new file will be created. All pre-existent content will be
+        overwritten by this operation.
+
+        Args:
+            path: Path to the file in the workload container.
+            content: the text content to be written to the file.
```
arturo-seijas: The `raises` section is missing from this docstring. `self.flask_container()` might throw an expcetion

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
+        """
+        self.flask_container().push(path, content, encoding="utf-8")
+
+    def pull_file(self, path: str) -> str | None:
+        """Retrieve the content of the given file from the flask workload container.
+
+        Args:
+            path: Path to the file in the workload container.
+
```
arturo-seijas: The `raises` section is missing from this docstring. `self.flask_container()` might throw an expcetion

sample_rock/sample-app/app.py
weiiwang01: Yes, the return value of a flask view function can not be `None`.

src/charm.py
weiiwang01: If the configuration has changed, it need to send a signal to the Gunicorn server to reload the configuration. The reload process is a little disruptive to the Gunicorn server, so I would like to eliminate the unnecessary reloads.

src/charm.py

src/charm.py

src/charm.py
weiiwang01: I think only the direct raises inside the function should be recorded in the docstring? The `RuntimeError` is raised indirectly in `self.flask_container`, so maybe we can omit that?

src/charm.py
weiiwang01: Same as above.

src/charm.py
gregory-schiano: Ok let me review the current version

sample_rock/sample-app/app.py
gregory-schiano: Perhaps better to return HTTP 204
`return "", http.HTTPStatus.NO_CONTENT`

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
+        self._charm_state = CharmState(charm_config=self.config)
```
gregory-schiano: I'd prefer passing the config elements the class needs and if there are too many, a named tuple as transitional object

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
+
     def config_service(self, event: ConfigChangedEvent) -> None:
         """Configure the flask pebble service layer.
 
         Args:
             event: the config-changed event that trigger this callback function.
         """
-        container = self.unit.get_container("flask-app")
+        container = self.flask_container(require_connected=False)
         if not container.can_connect():
```
gregory-schiano: Isn't it better to call `flask_container` with require_connected true and catch the exception to defer ?

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
+        self._charm_state = CharmState(charm_config=self.config)
+        self._webserver = GunicornWebserver(charm_state=self._charm_state)
         self.framework.observe(self.on.config_changed, self.config_service)
```
gregory-schiano: Can we rename the handler `_on_config_changed` as per specs ?
https://docs.google.com/document/d/1G62PosrObvmQY5KbxvqaxByojlhDxrmNtcbPS39YbaY/edit#bookmark=id.20k0qpk5irc9

src/charm.py
gregory-schiano: Can't we let the webserver service define if a reload is needed ?
This will let the service define how to determine if event triggered requires a signal to be sent and let the charm send the signal using pebble client

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
+        is_webserver_running = container.get_service(service_name).is_running()
+        if current_webserver_config != self._webserver.config:
+            self.push_file(webserver_config_path, self._webserver.config)
+            if is_webserver_running:
+                logger.info("gunicorn config changed, reloading")
+                container.send_signal(self._webserver.reload_signal, service_name)
         container.replan()
         self.unit.status = ActiveStatus()
 
+    def push_file(self, path: str, content: str) -> None:
```
gregory-schiano: I think it will be worth it to move this method and `pull_file` into a dedicated `pebble_helper` module

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
-        container = self.unit.get_container("flask-app")
+        container = self.flask_container(require_connected=False)
         if not container.can_connect():
             event.defer()
             return
+
+        service_name = "flask-app"
         container.add_layer("flask-app", self.flask_layer(), combine=True)
+        webserver_config_path = str(self._webserver.config_path)
+        current_webserver_config = self.pull_file(webserver_config_path)
```
gregory-schiano: `pull_file` will return `None` if the file doesn't exist shouldn't we expect the file to be available at that time ? And start generating the file in the `install` event ?
This will avoid relying on the config_change event being triggered at startup

src/charm_state.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""This module defines the CharmState class which represent the state of the Flask charm."""
```
gregory-schiano: s/represent/represents

src/charm_state.py
```diff
@@ -0,0 +1,96 @@
...
+class CharmState:
+    """Represents the state of the Flask charm."""
+
+    def __init__(self, charm_config: typing.Mapping[str, str]):
+        """Initialize a new instance of the CharmState class.
+
+        Args:
+            charm_config: Charm configurations of the charm instance associated with this state.
+        """
+        self._charm_config = charm_config
```
gregory-schiano: We should not store all the charm config in the State, we only take the elements we need

src/charm_state.py
```diff
@@ -0,0 +1,96 @@
...
+        keepalive = self._charm_config.get("webserver_keepalive")
+        timeout = self._charm_config.get("webserver_timeout")
+        workers = self._charm_config.get("webserver_workers")
+        threads = self._charm_config.get("webserver_threads")
+        return WebserverConfig(
+            workers=int(workers) if workers is not None else None,
+            threads=int(threads) if threads is not None else None,
+            keepalive=datetime.timedelta(seconds=int(keepalive))
+            if keepalive is not None
+            else None,
```
gregory-schiano: Why not simply keepalive then ?

gregory-schiano (commented):
I stopped before reviewing the unittests, I'll resume the review later

tests/integration/test_charm.py
```diff
@@ -41,3 +26,19 @@ async def test_flask_is_up(flask_app, get_unit_ips):
...
+):
+    """
+    arrange: build and deploy the flask charm, and change the gunicorn timeout configuration.
+    act: send long-running requests to the flask application managed by the flask charm.
+    assert: the gunicorn should restart the worker if the request duration exceeds the timeout.
+    """
+    for unit_ip in await get_unit_ips(flask_app.name):
+        assert requests.get(f"http://{unit_ip}:8000/sleep?duration=6", timeout=10).ok
+        with pytest.raises(requests.ConnectionError):
+            requests.get(f"http://{unit_ip}:8000/sleep?duration=8", timeout=10)
```
gregory-schiano: Does this effectively checks that a worker is restarted ? What are we really testing here ?
If you want to test that the configuration is properly propagated to gunicorn, isn't it better to add an endpoint that reports the server config if that's possible ?

src/charm.py
```diff
@@ -24,22 +28,80 @@ def __init__(self, *args: typing.Any) -> None:
...
+        is_webserver_running = container.get_service(service_name).is_running()
+        if current_webserver_config != self._webserver.config:
+            self.push_file(webserver_config_path, self._webserver.config)
+            if is_webserver_running:
```
gregory-schiano: In the specs I specify that the generated configuration is checked using `gunicorn --check-config` before starting/reloading

Could you add this ?

tests/unit/conftest.py
gregory-schiano: Isn't the pebble client already mocked in harness ?

tests/unit/test_charm.py
```diff
@@ -19,8 +20,36 @@ def test_flask_pebble_layer(harness: Harness) -> None:
     assert flask_layer == {
         "override": "replace",
         "summary": "Flask application service",
-        "command": "python3 -m gunicorn --chdir /srv/flask/app app:app -b 0.0.0.0:8000",
+        "command": "python3 -m gunicorn -c /srv/flask/gunicorn.conf.py app:app",
```
gregory-schiano: Shouldn't `/srv/flask` and `app:app` be constants ?
Event perhaps the command line using templated string ?

tests/unit/test_charm.py
```diff
@@ -19,8 +20,36 @@ def test_flask_pebble_layer(harness: Harness) -> None:
...
+    arrange: start the flask charm and set flask-app container to be ready.
+    act: update gunicorn related charm configurations.
+    assert: gunicorn configuration file inside the flask app container should change accordingly.
+    """
+    harness.begin_with_initial_hooks()
+    harness.set_can_connect(harness.model.unit.containers["flask-app"], True)
+    harness.framework.reemit()
+    harness.update_config({"webserver_workers": 1})
+    assert mock_container_fs.get("/srv/flask/gunicorn.conf.py") == "\n".join(
+        ["bind = ['0.0.0.0:8000']", "chdir = '/srv/flask/app'", "workers = 1"]
```
gregory-schiano: Same as above for the binding address and the port

gregory-schiano (commented):
Finished review

src/charm.py
weiiwang01: In my opinion, passing every individual configuration option breaks boundary between the `FlaskCharm` and `CharmState`. The process of determining which configuration options make up the charm state should be an internal implementation detail of CharmState, separated from the concerns of FlaskCharm.

src/charm.py
weiiwang01: The `RuntimeError` is just a safe guard to avoid accidental invocation without checking. Since the `RuntimeError` can be raised by many other processes, it's a little unsafe to catch that.

src/charm.py

src/charm.py
weiiwang01: If we generate the file during the install event then check the file in the config-changed event, that requires us to assume that install event happens before the config-changed event, which is exactly the kind of thing I would like to avoid.

src/charm_state.py
weiiwang01: Updated, thanks!

src/charm_state.py
weiiwang01: As I mentioned before, the process of determining which configuration options make up the charm state should be an internal implementation detail of CharmState, separated from the concerns of FlaskCharm.

src/charm_state.py
weiiwang01: The typing conversion required by static checks, the charm config is typed with `Mapping[str, str]`.

tests/integration/test_charm.py
weiiwang01: This test probes the exact timeout duration (less than 8 and greater than 6) using the sleep endpoint.
There's no way to retrieve the Gunicorn config inside flask, so I have to be creative here.

src/charm.py
gregory-schiano: In my opinion the state should only know and get the information it needs, passing the whole charm config doesn't make sense for that as state will have information it doesn't need, doesn't care and shouldn't care.
In terms of design boundaries any object should not get access to any data or method it's not suppose to read or call (hence the concept of interface that only exposes what you need to expose).

@jdkandersson a topic to discuss together

src/charm.py
gregory-schiano: Well then we can have a custom exception, more specific raised by `flask_container`

src/charm.py
weiiwang01: If the webserver service can can determine if a reload is needed, the webserver service is not purely about the webserver anymore, it will be a web server pebble service. It will break the single-responsibility principle.

tests/unit/test_charm.py
weiiwang01: Do you mean like this?
```python
uwsgi_app_path = "app:app"
flask_base_dir = "/srv/flask"
command = f"python3 -m gunicorn -c {flask_base_dir}/gunicorn.conf.py {uwsgi_app_path}"
```

src/charm.py
weiiwang01: If the configuration is passed the the `CharmState` one by one. When the `CharmState` changed, for example, a new configuration was added, the interface of the `CharmState` has to change too to add the new configuration, this breaks the open-close principle.

src/charm.py
weiiwang01: I created a new method called `flask_container_can_connect` to check if the pebble is ready and removed the `require_connected` argument from `flask_container`.
I think it's too much to create an new entity (custom exception) for this function.

src/charm.py
weiiwang01: If I would like the move the `pull_file` and `push_file` into a new class, all the pebble APIs (replan, layer, signal) has to be moved to that class too. Then at that time, the new class will be an abstraction over the `ops.model.Container`. But the `ops.model.Container` is already an abstraction over the pebble client. This make the new class an abstraction over an abstraction, which increase the complexity of the code.

The `pull_file` and `push_file` is just here so I can align the pebble file interface with the Python mapping interface (`get` and `__setitem__`) to make the mocking simpler.

tests/unit/conftest.py
weiiwang01: Not the file and exec API.

src/charm.py
weiiwang01: Added, thanks!

src/charm.py
arturo-seijas: Maybe

src/charm.py
```diff
@@ -121,7 +147,7 @@ def flask_layer(self) -> dict:
                 "flask-app": {
                     "override": "replace",
                     "summary": "Flask application service",
-                    "command": self._webserver.command,
+                    "command": shlex.join(self._webserver.command),
```
arturo-seijas, line 106: A little picky but I think it would be more cohesive to return the string resulting from `shlex.join` in `self._webserver.command`

weiiwang01: Yes, there's another method called `GunicornWebserver.check_config_command` that is used by `ops.model.Container.exec`, which requires a list as its argument. To maintain consistency between the `check_config_command` and `command` methods, I decided to let both of them to return a list of strings. Since converting from a list of strings to a single string is simpler than the other way around.

tests/unit/conftest.py
gregory-schiano: Ok that's probably a topic to discuss with the Juju team

src/charm.py
gregory-schiano: I don't agree, defining if a reload is needed is purely the webserver responsibility, some other webserver could require a restart in other condition therefore we need an interface to specify if a reload is needed.
I don't expect the webserver service to effectively do the reload, but to determine if a reload is needed or not depending on the state changes.
Therefore having a "should_reload" or "must_reload" method in the webserver service that will implement the logic and return true if a reload is needed is the way to go.
And then the charm execute the reload logic through Pebble

src/charm.py
gregory-schiano: Well it does make sense if the charmstate changes, it requires changes to its interface if that information must be exposed / consumed.
After discussing with David we came to a proposal that has been reflected in the charm complexity spec see

src/charm.py
gregory-schiano: I agree that right now it's not useful, it will be for a later discussion if the charm.py grows

src/charm.py
gregory-schiano: Yes indeed you're right, better not assume event ordering and config-changed is documented as being triggered during deploy and upgrade scenario so we can rely on that event to be triggered when we need it

src/charm_state.py
gregory-schiano: See the other discussion, I close this comment to avoid duplicates

src/charm_state.py
gregory-schiano: Raaah that makes me crazy to have to write useless code for a tool :|

tests/unit/test_charm.py
gregory-schiano: Yeap something like this and shared with the webserver service

src/charm.py
arturo-seijas: Understood. Thx

src/charm.py
weiiwang01: Indeed some web server require a restart in other condition, but if the `FlaskCharm` doesn't know when this "other condition" would happen, the `FlaskCharm` wouldn't have the chance to call the `should_reload` or `must_reload`. We can't just call the `should_reload` and `must_reload` between every line. This is why the `FlaskCharm` must know when the web server should reload in order to reload the web server.

tests/integration/test_charm.py
weiiwang01: Sorry, I misinterpreted your words before. The `webserver_timeout` is the only webserver configuration that's easy to probe from the server. Since the Flask can not access the Gunicorn configuration, this is the only way we have.

tests/unit/test_charm.py
weiiwang01: Updated, but since only the `"/srv/flask"` is shared between two test cases, I just extracted that one, thanks!

tests/unit/test_charm.py
weiiwang01: Same as above.

tests/integration/test_charm.py
jdkandersson: @gregory-schiano could you look at this just to double check please?

src/charm.py
weiiwang01: I am updating the CharmState to accept individual charm configuration values instead of the entire configuration. However, I have encountered another problem in the operator framework testing module. The charm class is initialized only once for the entire test case. So, when handling multiple events in unit test cases, the charm constructor is called only once. This causes CharmState to become stalled with old data since the data update happens in the charm class constructor. I am not sure whether this is a bug or a feature.

tests/integration/test_charm.py
gregory-schiano: I think there could be a way using https://docs.gunicorn.org/en/latest/custom.html#custom-application

Doing something like: 
```python
from flask import Flask, jsonify
from gunicorn.app.base import BaseApplication

app = Flask(__name__)

class CustomGunicornApplication(BaseApplication):
    def __init__(self, app, options=None):
        self.options = options or {}
        self.application = app
        super().__init__()

    def load_config(self):
        config = {key: value for key, value in self.options.items()
                  if key in self.cfg.settings and value is not None}
        for key, value in config.items():
            self.cfg.set(key.lower(), value)

        # Store the Gunicorn config in the Flask app config
        self.application.config['gunicorn_config'] = config

    def load(self):
        return self.application

@app.route('/')
def hello():
    return 'Hello, World!'

@app.route('/config')
def config():
    gunicorn_config = app.config.get('gunicorn_config', {})
    return jsonify(gunicorn_config)

if __name__ == '__main__':
    options = {
        'bind': '%s:%s' % ('127.0.0.1', '8000'),
        'workers': 4,
    }
    CustomGunicornApplication(app, options).run()
```

Would probably require to inherit from: https://github.com/benoitc/gunicorn/blob/master/gunicorn/app/wsgiapp.py instead of BaseApplication

It sounds a bit complex, but should make it

sample_rock/sample-app/app.py
gregory-schiano: @weiiwang01 any feedback ?

src/charm.py
gregory-schiano: I'm not telling that the charm must not know when the web server requires a restart
I'm just telling that it's not the responsibility to define the business logic that decides if the webserver requires a restart.
All the charm needs to know if "do I need to reload/restart the webserver", if yes, then I do it because I manage pebble

It's not the charm responsibility to know that if condition X and Y are met, then a reload is needed because these conditions depends on the webserver

So what I what is that the charm does something like:
```python
if self._webserver.should_reload(current_webserver_config):
   self.reload_webserver()
```

And on the webserver module:
```python
def should_reload(current_webserver_config: dict[str, str]): bool
   return self.config != current_webserver_config
```

Therefore if another webserver needs to look only at the nb of workers it can do instead:
```python
def should_reload(current_webserver_config: dict[str, str]): bool
   return self.config['workers'] != current_webserver_config['workers'] && current_webserver_config['workers'] > 0
```

weiiwang01: The problem is that the charm has to know when the webserver needs to reload to know which place should the snippet `if self._webserver.should_reload(current_webserver_config)...` be put inside the charm. And if the charm already know when the webserver needs to reload, the charm can just reload by it self.

tests/integration/test_charm.py
weiiwang01: This script need a different command `python3 app.py` to start the web server, if you are using the regular Gunicorn command `python3 -m gunicorn app:app` to start the server, the section in `if __name__ == '__main__':` is not executed.

src/charm.py
gregory-schiano: Let's make a call, I think we don't understand each other :D
What you say is not incompatible with what I say:
It's the charm that does the reload in my example, it's the charm that decides when the reload is done in my example, but it's the webserver service that knows if a reload is needed

weiiwang01: I updated the code to use webserver service to decide the reload based on your comment, thanks!

tests/integration/test_charm.py
gregory-schiano: Indeed, that requires to change the command.
An alternative would be to add:
```python
def when_ready(server):
    app = server.app.callable

    # Extract Gunicorn configuration
    gunicorn_config = {key: value for key, value in server.cfg.settings.items()
                       if value.value is not None and value.value != ''}
    
    # Set Gunicorn configuration in the Flask app
    app.config['gunicorn_config'] = gunicorn_config
```

To the gunicorn config and then an endpoint like:
```python
@app.route('/config')
def config():
    gunicorn_config = app.config.get('gunicorn_config', {})
    return jsonify(gunicorn_config)
```

The question is now, how do we add this to the gunicorn configuration

gregory-schiano: Ok got something that works
We can systematically add this to the gunicorn configuration
```python
import json
import os


class CustomJSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if callable(obj):
            return f"<function: {obj.__name__}>"
        if hasattr(obj, "__dict__"):
            return obj.__dict__
        return f"<non-serializable: {type(obj).__name__}>"


def post_worker_init(server):
    if os.getenv("ENABLE_CONFIG_HOOK", "false").lower() == "true":
        app = server.app.callable

        # Extract Gunicorn configuration
        gunicorn_config = {
            key: value
            for key, value in server.cfg.settings.items()
            if value.value is not None and value.value != ""
        }

        gunicorn_config_serializable = json.loads(
            json.dumps(gunicorn_config, cls=CustomJSONEncoder)
        )

        # Set Gunicorn configuration in the Flask app
        app.config["gunicorn_config"] = gunicorn_config_serializable
```

Then either exposing the ENABLE_CONFIG_HOOK through a charm config
Or exposing an EXTRA_ENV configuration

This is specific to the integration test and not very straightforward
What's you opinion @jdkandersson ?

weiiwang01: Config hook is very specific to Gunicorn and will not apply to other WSGI servers.

I believe the current implementation is sufficient to test the timeout configuration value. I have added some tests that use concurrent requests to test the threads and workers configuration values. I also repeated each experiment three times to increase the confidence level, or I can add more repetitions if you want.

I think the current integration tests are "scientific" enough.

src/webserver.py
```diff
@@ -64,36 +75,76 @@ def command(self) -> list[str]:
...
 
         Returns:
-            bool: True if the 'current_config' is different from the desired configuration and
-                a reload is need to apply the new configuration, False otherwise.
+            ExecResult: An `ExecResult` object representing the result of the command execution.
         """
-        return self.config != current_config
+        container = self._flask_container
+        exec_process = container.exec(command, user="flask", group="flask")
+        try:
```
arturo-seijas, line 121: Why not raising this exception directly and handling it within the caller?

weiiwang01: I think that the command execution will be used multiple times in this class. 
So I put the exception handling inside to avoid repeating the exception handling section.

pyproject.toml
```diff
@@ -33,7 +33,11 @@ select = ["E", "W", "F", "C", "N", "R", "D", "H"]
 # Ignore D107 Missing docstring in __init__
 ignore = ["W503", "E501", "D107"]
 # D100, D101, D102, D103: Ignore missing docstrings in tests
-per-file-ignores = ["tests/*:D100,D101,D102,D103,D104,D205,D212,D415,DCO020,DCO030"]
+per-file-ignores = [
+    "tests/*:D100,D101,D102,D103,D104,D205,D212,D415,DCO020,DCO030,DCO050",
+    "src/webserver.py:DCO060", # false alarms for classes with property functions
```
jdkandersson: What is the reason for this? Do you think we should document properties in the docstring?

pyproject.toml
```diff
@@ -33,7 +33,11 @@ select = ["E", "W", "F", "C", "N", "R", "D", "H"]
...
+    "src/charm_state.py:DCO060,DCO063" # false alarms for classes with property functions
```
jdkandersson: Similar as above

src/charm_state.py
```diff
@@ -0,0 +1,116 @@
...
+import datetime
+import pathlib
+
+from ops.charm import CharmBase
+
+from charm_types import WebserverConfig
+
+
+class CharmState:
+    """Represents the state of the Flask charm."""
```
jdkandersson: Do you think it is worth documenting the `@property` similar to attributes?

src/charm_state.py
```diff
@@ -0,0 +1,116 @@
...
+        webserver_workers: int | None = None,
+        webserver_threads: int | None = None,
+        webserver_keepalive: int | None = None,
+        webserver_timeout: int | None = None,
+    ):
+        """Initialize a new instance of the CharmState class.
+
+        Args:
+            webserver_workers: The number of workers to use for the web server,
+                or None if not specified.
```
jdkandersson, line 36: Is it worth noting that it defaults to `None`? The function signature already shows this?

src/charm_state.py
```diff
@@ -0,0 +1,116 @@
...
+
+        Return:
+            The CharmState instance created by the provided charm.
+        """
+        keepalive = charm.config.get("webserver_keepalive")
+        timeout = charm.config.get("webserver_timeout")
+        workers = charm.config.get("webserver_workers")
+        threads = charm.config.get("webserver_threads")
+        return cls(
+            webserver_workers=int(workers) if workers is not None else None,
```
jdkandersson, line 64: What is the reason for the `is None` check please? It seems that `__init__` is fine with taking `None` values?

src/webserver.py
```diff
@@ -0,0 +1,151 @@
...
+            setting_value = getattr(self._charm_state.webserver_config, setting)
+            if setting_value is None:
+                continue
+            setting_value = (
+                setting_value
+                if isinstance(setting_value, int)
+                else int(setting_value.total_seconds())
+            )
+            config_entries.append(f"{setting} = {setting_value}")
+        return "\n".join(config_entries)
```
jdkandersson: Do you think this should be done using jinja? It would make it easier to see what the final configuration looks like?

tests/integration/test_charm.py
```diff
@@ -28,17 +30,67 @@ async def test_flask_is_up(
         assert "Hello, World!" in response.text
 
 
-@pytest.mark.parametrize("update_config", [{"webserver_timeout": 7}], indirect=True)
+@pytest.mark.parametrize(
+    "update_config, timeout",
+    [({"webserver_timeout": 7}, 7), ({"webserver_timeout": 5}, 5), ({"webserver_timeout": 3}, 3)],
+    indirect=["update_config"],
+    ids=["timeout-7", "timeout-5", "timeout-3"],
```
jdkandersson: Do you think it would be clearer to put the `id`s with the definition for each test? E.g.:

```python
[pytest.param({"webserver_timeout": 7}, 7, id="timeout-7"), ...]
```

tests/integration/test_charm.py
```diff
@@ -28,17 +30,67 @@ async def test_flask_is_up(
...
+    safety_timeout = timeout + 3
     for unit_ip in await get_unit_ips(flask_app.name):
-        assert requests.get(f"http://{unit_ip}:8000/sleep?duration=6", timeout=10).ok
+        assert requests.get(
+            f"http://{unit_ip}:8000/sleep?duration={timeout - 1}", timeout=safety_timeout
+        ).ok
         with pytest.raises(requests.ConnectionError):
-            requests.get(f"http://{unit_ip}:8000/sleep?duration=8", timeout=10)
+            requests.get(
+                f"http://{unit_ip}:8000/sleep?duration={timeout + 1}", timeout=safety_timeout
```
jdkandersson, line 63: What is the reason for having multiple of these checks?

tests/integration/test_charm.py
```diff
@@ -28,17 +30,67 @@ async def test_flask_is_up(
...
+    model: juju.model.Model,
+    flask_app: Application,
+    get_unit_ips,
+):
+    """
+    arrange: build and deploy the flask charm.
+    act: deploy the ingress, configure it and relate it to the charm.
+    assert: requesting the charm through traefik should return a correct response
+    """
+    traefik_app = await model.deploy("traefik-k8s", trust=True)
```
jdkandersson: Should this be a fixture?

jdkandersson: I guess up to line 87

tox.ini
```diff
@@ -33,7 +33,7 @@ deps =
     isort
 commands =
     isort {[vars]all_path}
-    black {[vars]all_path}
+    black {[vars]all_path} --target-version=py311
```
jdkandersson: Should we be targeting python 3.10 since that is what comes with 22.04?

pyproject.toml
weiiwang01: I added the `Attrs` section to the class docstring, this ignore has been removed, thanks!

pyproject.toml
weiiwang01: Same as above.

src/charm_state.py
weiiwang01: Added, thanks!

src/charm_state.py
weiiwang01: Yes, I think the function signature already states that. And I think the second sentence implies that `None` is the default value.

src/charm_state.py
weiiwang01: This is just for type checks since the `charm.config` is typed with `Map[str, str | None]`

src/webserver.py
weiiwang01: I found it difficult to use jinja to generate the configuration file since most of the configuration file is dynamic with very little static content. This makes it not very suited for template.

tests/integration/test_charm.py
weiiwang01: The two requests (timeout+1, timeout-1) is to probe the upper bound and lower bound for the timeout value.

tox.ini
weiiwang01: Yes, you are correct. Updated, thanks!

tests/integration/test_charm.py
weiiwang01: Doesn't know that before, updated, thanks!

tests/integration/test_charm.py
weiiwang01: I have relocated this part to the deploy fixture, which can save us some time. Thanks!

.licenserc.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - '**/.jujuignore'
+    - '**/.gitignore'
+    - '**/.licenserc.yaml'
+    - '**/.trivyignore'
+    - '**/CODEOWNERS'
+    - '**/icon.svg'
+    - '**/LICENSE'
+    - '**/trivy.yaml'
+    - 'lib/charms/traefik_k8s/v1/ingress.py'
+  comment: on-failure
```
arturo-seijas: Missing blank line at EOF

src/webserver.py
```diff
@@ -0,0 +1,156 @@
...
+            WebserverConfigInvalidError: If the web server configuration check fails.
+        """
+        webserver_config_path = str(self._config_path)
+        try:
+            current_webserver_config = self._flask_container.pull(webserver_config_path)
+        except PathError:
+            current_webserver_config = None
+        self._flask_container.push(webserver_config_path, self._config)
+        if current_webserver_config == self._config:
+            return
```
amandahla, line 145: Is this just to check if the current and new configurations are the same?  Wouldn't be useful to log a message that no action will be taken if this is the case?

tests/integration/test_charm.py
```diff
@@ -26,35 +32,55 @@ async def test_flask_is_up(flask_app, get_unit_ips):
...
+@pytest.mark.usefixtures("update_config")
+async def test_flask_webserver_timeout(
+    flask_app: Application,
+    get_unit_ips: typing.Callable[[str], typing.Awaitable[tuple[str, ...]]],
+    timeout,
+):
+    """
+    arrange: build and deploy the flask charm, and change the gunicorn timeout configuration.
+    act: send long-running requests to the flask application managed by the flask charm.
+    assert: the gunicorn should restart the worker if the request duration exceeds the timeout.
```
amandahla, line 54: For each unit IP the test checks if the request is OK for timeout-1 and then checks if raises an exception for timeout+1. But how is verifying that the worker was restarted? Should it test again with timeout-1?

.licenserc.yaml
weiiwang01: Added, thanks!

tests/integration/test_charm.py
weiiwang01: The default timeout is 30 seconds, the change of the timeout value implies that the server has been reloaded.

src/webserver.py
weiiwang01: This will happen every time there's a config-changed event, which means it will happen quite frequently. I feel it might be a little bit redundant to notify the user that configuration hasn't been changed. The flask charm prints a log message when the configuration changed, do you think that's good enough?

tests/integration/test_charm.py
jdkandersson: Sounds reasonable to me, we can always use the suggestion from Gregory if needed

src/charm_state.py
jdkandersson: Looks good, thanks!

src/charm_state.py
jdkandersson: Potentially worth adding a comment to explain, up to you

src/webserver.py
jdkandersson: Would it be possible to provide an example of what the jinja template would look like in this case please? It would be good to give a few people the chance to see it because based on the code, it is hard to visualise what it would look like right now

tests/integration/test_charm.py
```diff
@@ -4,6 +4,8 @@
 
 """Integration tests for Flask charm."""
 
+# pylint: disable=too-many-arguments
```
jdkandersson, line 8: Should this include an explanatory comment why this disable is needed?

weiiwang01: Yes, it's caused by pytest fixtures. I have added that comment.

src/webserver.py
weiiwang01: Here's a simple example using jinja2:
```python
from jinja2.sandbox import SandboxedEnvironment
import pathlib
import datetime

template_content = """\
bind = ['0.0.0.0:{{ flask_port }}']
chdir = {{ repr(str(flask_dir.absolute())) }}
{%- for setting in ("workers", "threads", "keepalive", "timeout") -%}
    {%- if webserver_config.get(setting) %}
{{ setting }} = {{ render_setting_value(webserver_config.get(setting)) }}
    {%- endif %}
{%- endfor -%}
"""

env = SandboxedEnvironment()
env.globals["repr"] = repr
env.globals["str"] = str
template = env.from_string(template_content)

print(
    template.render(
        flask_port=8000,
        flask_dir=pathlib.Path("/srv/flask/app"),
        render_setting_value=lambda v: v if isinstance(v, int) else v.total_seconds(),
        webserver_config={"timeout": datetime.timedelta(seconds=1), "workers": 10},
    )
)
```

jdkandersson: I kind of prefer this, I can tell what the final file will look like much more easily. Any thoughts @gregory-schiano?

jdkandersson: What is the reason for needing `repr` and `str` around `flask_dir.absolute()`? I guess `flask_dir=repr(str(pathlib.Path("/srv/flask/app")))` in `template.render` could work as well? That way `env.globals` doesn't need to have `str` and `repr`?

weiiwang01: I think the problem of this jinja2 template is that almost everything is enclosed inside the `{% %}` and `{{ }}`, what make it less legible.

jdkandersson: The UI developer in me is used to that 🙂

jdkandersson: it is very common to have that in HTML rendering and other UI libraries

weiiwang01: > What is the reason for needing repr and str around flask_dir.absolute()? I guess flask_dir=repr(str(pathlib.Path("/srv/flask/app"))) in template.render could work as well? That way env.globals doesn't need to have str and repr?

I think format and rendering of the content is the responsibility of template, so I put the formatting inside the template.

jdkandersson: > I think format and rendering of the content is the responsibility of template, so I put the formatting inside the template.

That is a fair point

gregory-schiano: I don't feel like using a template engine for a simple `key = value` file worth it.
Templating is always risky in a project as it could lead to some mixed-ugly things, no template will avoid this.

But I'm wondering if the code could be simplified by considering every value as a string and preparing its value representations before and merge with the additional parameters we put (bind and chdir)

gregory-schiano:
Can you add a description to the PR please ?

src/webserver.py
jdkandersson: maybe instead of using a templating engine we can use a nicely formatted multi-line f-string that looks much like the template for jinja above?

gregory-schiano: I'm in :)

weiiwang01: The problem of using multi-line strings is that it doesn't change many things.

```python
def _config(self) -> str:
    config_entries = []
    settings = ("workers", "threads", "keepalive", "timeout")
    for setting in settings:
        setting_value = getattr(self._charm_state.webserver_config, setting)
        if setting_value is None:
            continue
        setting_value = (
            setting_value
            if isinstance(setting_value, int)
            else int(setting_value.total_seconds())
        )
        config_entries.append(f"{setting} = {setting_value}")
    new_line = "\n"
    config = textwrap.dedent(
        f"""\
        bind = ['0.0.0.0:{self._charm_state.flask_port}
        chdir = {repr(str(self._charm_state.flask_dir.absolute()))}
        {new_line.join(config_entries)}"""
    )
    return config
```

src/webserver.py
amandahla: I see, you're right, it would be redundant.

src/webserver.py
jdkandersson: To me that looks good, at least the structure is a bit clearer

weiiwang01: Okay, updated, thanks

jdkandersson (approved):

src/charm_types.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
yanksyoon, line 1: I think this shebang line could be removed

yanksyoon (approved):
Minor comment, thank you!

src/charm_types.py
weiiwang01: I will remove that in the next pull request, thanks!

PR #4, opened 2023-04-17
Fix rockcraft and add instructions in CONTRIBUTING

nrobinaubertin:
Fix the permissions of the `/srv/flask` directory inside the rockcraft image.

### DIFF
--- a/sample_rock/rockcraft.yaml
+++ b/sample_rock/rockcraft.yaml
@@ -38,3 +38,8 @@ parts:
       sample-app: srv/flask/app
     prime:
       - srv/flask/app
+    permissions:
+      - owner: 2000
+        group: 2000
+      - path: srv/flask/
+        mode: 755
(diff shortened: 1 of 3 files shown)


sample_rock/rockcraft.yaml
```diff
@@ -38,3 +38,8 @@ parts:
       sample-app: srv/flask/app
     prime:
       - srv/flask/app
+    permissions:
+      - owner: 2000
+        group: 2000
```
gregory-schiano, line 43: Aren't the permissions kept from the overlay script where there's a chown ?

gtrkiller (approved):
LGTM

sample_rock/rockcraft.yaml
nrobinaubertin: The `flask-app` part is dumping files from the host filesystem into `/srv/flask/app`. Those new files are not affected by the previous `chown`

edit: actually I need to test that. There were potentially two issues: permissions and owner.

nrobinaubertin: After testing, the "owner, group" part is required.

jdkandersson (approved):

PR #6, opened 2023-04-21
Add ingress relation support and corresponding integration test

nrobinaubertin:
Adds the `ingress` interface support and the corresponding integration test.

We take as granted that the flask app will always be on port 8000.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,6 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
 # Testing tools configuration
 [tool.coverage.run]
 branch = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,9 @@ skips = ["*/*test.py", "*/test_*.py"]
 
 [tool.mypy]
 ignore_missing_imports = true
-explicit_package_bases = true
-namespace_packages = true
 check_untyped_defs = true
+disallow_untyped_defs = true
+
+[[tool.mypy.overrides]]
+module = "tests.*"
+disallow_untyped_defs = false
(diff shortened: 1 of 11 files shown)


gregory-schiano:
Can you add a PR description ?

src/charm.py
```diff
@@ -25,6 +26,12 @@ def __init__(self, *args: typing.Any) -> None:
         """
         super().__init__(*args)
         self.framework.observe(self.on.config_changed, self.config_service)
+        self.ingress = IngressPerAppRequirer(
+            self,
+            port=8000,
```
gregory-schiano: I think we should define a constant for the port, it's repeated

src/charm.py
```diff
@@ -25,6 +26,12 @@ def __init__(self, *args: typing.Any) -> None:
...
+            host=f"{self.app.name}-endpoints.{self.model.name}.svc.cluster.local",
```
gregory-schiano, line 37: Can you add a comment for this that point to the bug that forces us to use this trick ?

pyproject.toml
```diff
@@ -42,3 +42,9 @@ exclude_dirs = ["/venv/"]
 
 [tool.bandit.assert_used]
 skips = ["*/*test.py", "*/test_*.py"]
+
+[tool.mypy]
+ignore_missing_imports = true
```
arturo-seijas, line 50: I think this doesn't match what's defined in https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#type-hints

src/charm.py
nrobinaubertin: Added

src/charm.py
nrobinaubertin: Fixed

pyproject.toml
nrobinaubertin: I sync'ed it with our guide but I still need to keep ignore_missing_imports for the ingress library

arturo-seijas (approved):

amandahla (approved):
LGTM

PR #7, opened 2023-04-27
Auto-update charm libraries

arturo-seijas:


.github/workflows/auto_update_libs.yaml
```diff
@@ -0,0 +1,10 @@
+name: Auto-update charm libraries
```
amandahla, line 1: I've just realized something: do you think it would be useful to add a badge for this workflow?

arturo-seijas: Not sure if it makes sense. What do you think?

nrobinaubertin (approved):

.github/workflows/auto_update_libs.yaml
arturo-seijas: Anyway, I'd say that's out of the scope of this PR, but I'm happy to discuss it further :)

gtrkiller (approved):
LGTM

PR #9, opened 2023-05-10
Simplify rockfile

nrobinaubertin:
I tried to simplify the rockfile with this PR.

I removed the non-root user that we added since I believe a streamlined way to do what we want could come later with this PR: https://github.com/canonical/rockcraft/pull/233

Additional thoughts:
- We can't easily use a `pyproject.toml` for now. `requirements.txt` is the way to go
- if the `requirements.txt` file is not present, the rockcraft will fail
- We decided to have flask inside `requirements.txt` as shown in [flask's guide](https://flask.palletsprojects.com/en/2.3.x/installation/)
- gunicorn is installed system-wide but as a pypi module (in order to have a more up to date package than through apt)
- We want to use `python3` in the charm and let the [PATH](https://github.com/canonical/pebble/blob/dbda12237fef3c4d2739824fce7fa65ba1dad76a/internal/overlord/cmdstate/request.go#L80) decide of the executable (to adapt to the eventual user custom OCI images). There's currently [a bug](https://github.com/canonical/rockcraft/issues/242) that forces us to remove `/usr/bin/python3` for that to work properly.


weiiwang01:
I have complex feeling about the builtin Python plugin. In one hand, this is provided by default, so it's very clean to use it. In another hand, I am not very agree with the Python plugin's approach to install Python dependencies.

jdkandersson:
> I have complex feeling about the builtin Python plugin. In one hand, this is provided by default, so it's very clean to use it. In another hand, I am not very agree with the Python plugin's approach to install Python dependencies.

Since it is provided by the tool I think we should use the plugin. We can give feedback on the plugin to @sergiusens

jdkandersson (dismissed):

jdkandersson:
A PR to take a look at @gregory-schiano

weiiwang01:
> A PR to take a look at @gregory-schiano



> > I have complex feeling about the builtin Python plugin. In one hand, this is provided by default, so it's very clean to use it. In another hand, I am not very agree with the Python plugin's approach to install Python dependencies.
> 
> Since it is provided by the tool I think we should use the plugin. We can give feedback on the plugin to @sergiusens

Copying one of the emails between Niels and me for the context:

Python plugin in *craft uses a very unorthodox way to install Python dependencies.
Packages are installed in `/lib/python3.10/site-packages`, which is not a system or user site directory.
And packages in `/lib/python3.10/site-packages` are put in the Python path by `/pyvenv.cfg` in the root directory.
So, `/bin/python3` can lookup packages correctly but not `/usr/bin/python3`.

jdkandersson:
> Copying one of the emails between Niels and me for the context:
> 
> Python plugin in *craft uses a very unorthodox way to install Python dependencies. Packages are installed in `/lib/python3.10/site-packages`, which is not a system or user site directory. And packages in `/lib/python3.10/site-packages` are put in the Python path by `/pyvenv.cfg` in the root directory. So, `/bin/python3` can lookup packages correctly but not `/usr/bin/python3`.

With all that noted, should we use the plugin?

nrobinaubertin:
> > Copying one of the emails between Niels and me for the context:
> > Python plugin in *craft uses a very unorthodox way to install Python dependencies. Packages are installed in `/lib/python3.10/site-packages`, which is not a system or user site directory. And packages in `/lib/python3.10/site-packages` are put in the Python path by `/pyvenv.cfg` in the root directory. So, `/bin/python3` can lookup packages correctly but not `/usr/bin/python3`.
> 
> With all that noted, should we use the plugin?

I think the key is to understand why the plugin does things this way. I'm going to ask around.

weiiwang01:
> > > Copying one of the emails between Niels and me for the context:
> > > Python plugin in *craft uses a very unorthodox way to install Python dependencies. Packages are installed in `/lib/python3.10/site-packages`, which is not a system or user site directory. And packages in `/lib/python3.10/site-packages` are put in the Python path by `/pyvenv.cfg` in the root directory. So, `/bin/python3` can lookup packages correctly but not `/usr/bin/python3`.
> > 
> > 
> > With all that noted, should we use the plugin?
> 
> I think the key is to understand why the plugin does things this way. I'm going to ask around.

The plugin is designed for `snap`, in that case, it is suitable since the content of snap is extracted to a snap directory. But in the case of OCI image, this plugin **does** work. But the image created by this plugin is a little bit unusual.

src/charm.py
```diff
@@ -64,11 +64,9 @@ def flask_layer(self) -> dict:
                     "override": "replace",
                     "summary": "Flask application service",
                     "command": (
-                        "python3 -m gunicorn --chdir /srv/flask/app app:app"
+                        "/bin/python3 -m gunicorn --chdir /srv/flask/app app:app"
```
arturo-seijas: Are you running this command as root now? The idea behind running as non root is to enforce the principle of least privilege. The containers do not totally isolate the process from the rest of processes running on the host

nrobinaubertin: I removed the non-root user that we added since I believe a streamlined way to do what we want could come later with this PR: https://github.com/canonical/rockcraft/pull/233

sample_rock/rockcraft.yaml
```diff
@@ -11,37 +11,21 @@ platforms:
...
-      chmod 777 $CRAFT_OVERLAY/tmp
-      groupadd -R $CRAFT_OVERLAY --gid 2000 flask
-      useradd -R $CRAFT_OVERLAY --system --gid 2000 --uid 2000 --home /srv/flask flask
   flask-dependencies:
+    plugin: python
+    source: .
     stage-packages:
-      - python3-dev
+      - python3-venv
       - python3-gunicorn
```
gregory-schiano: I think that in the latest discussion of 12-factor we decided that we will use pip for gunicorn and flask because the deb packages are lagging behind and that could cause pb
So I think we can remove this and install flask and gunicorn through pip install

gregory-schiano:
> > > > Copying one of the emails between Niels and me for the context:
> > > > Python plugin in *craft uses a very unorthodox way to install Python dependencies. Packages are installed in `/lib/python3.10/site-packages`, which is not a system or user site directory. And packages in `/lib/python3.10/site-packages` are put in the Python path by `/pyvenv.cfg` in the root directory. So, `/bin/python3` can lookup packages correctly but not `/usr/bin/python3`.
> > > 
> > > 
> > > With all that noted, should we use the plugin?
> > 
> > 
> > I think the key is to understand why the plugin does things this way. I'm going to ask around.
> 
> The plugin is designed for `snap`, in that case, it is suitable since the content of snap is extracted to a snap directory. But in the case of OCI image, this plugin **does** work. But the image created by this plugin is a little bit unusual.

Well the plugin should be designed and working for *craft tools. I think that we should use the rockcraft way and not using workarounds if something doesn't work the intended way.
Let's see what the rockcraft team has to tell about it

gregory-schiano:
> > > > > Copying one of the emails between Niels and me for the context:
> > > > > Python plugin in *craft uses a very unorthodox way to install Python dependencies. Packages are installed in `/lib/python3.10/site-packages`, which is not a system or user site directory. And packages in `/lib/python3.10/site-packages` are put in the Python path by `/pyvenv.cfg` in the root directory. So, `/bin/python3` can lookup packages correctly but not `/usr/bin/python3`.
> > > > 
> > > > 
> > > > With all that noted, should we use the plugin?
> > > 
> > > 
> > > I think the key is to understand why the plugin does things this way. I'm going to ask around.
> > 
> > 
> > The plugin is designed for `snap`, in that case, it is suitable since the content of snap is extracted to a snap directory. But in the case of OCI image, this plugin **does** work. But the image created by this plugin is a little bit unusual.
> 
> Well the plugin should be designed and working for *craft tools. I think that we should use the rockcraft way and not using workarounds if something doesn't work the intended way. Let's see what the rockcraft team has to tell about it

Note the following in the rockcraft repo:
https://github.com/canonical/rockcraft/blob/main/rockcraft/plugins/python_plugin.py

```
- The shebang in console scripts is hardcoded to "#!/bin/python3". In fact,
      every use of Python in the resulting image should be via /bin/python3.
```

We need to understand the reasoning behind as this could have consequences

sample_rock/rockcraft.yaml
nrobinaubertin: That will surely simplify things

gtrkiller (dismissed):
LGTM

sample_rock/rockcraft.yaml
nrobinaubertin: Fixed this

jdkandersson (dismissed):

yanksyoon (dismissed):
LGTM!

jdkandersson (dismissed):

src/webserver.py
```diff
@@ -79,7 +79,7 @@ def command(self) -> list[str]:
             The command to start the Gunicorn web server.
         """
         return [
-            "python3",
+            "/bin/python3",
```
gregory-schiano: I don't think it's a good idea, if we want to keep supporting custom OCI image for users, hardcoding the path here will cause troubles
I see 2 options:
* Appending "/bin" to the $PATH so that python3 will be found
* Providing a configuration option

sample_rock/rockcraft.yaml
```diff
@@ -11,37 +11,19 @@ platforms:
...
     source: .
-    plugin: nil
-    overlay-script: |
-      mkdir -p $CRAFT_OVERLAY/srv/flask/.local/lib/python3.10/site-packages
-      pip install -r $CRAFT_PART_SRC_WORK/sample-app/requirements.txt --target $CRAFT_OVERLAY/srv/flask/.local/lib/python3.10/site-packages
-      chown 2000:2000 -R $CRAFT_OVERLAY/srv/flask/
+    stage-packages:
+      - python3-venv
+    python-packages:
+      - gunicorn
```
gregory-schiano, line 20: Could you also add flask in the systematically installed packages ?

nrobinaubertin: We don't leave the choice of the flask version to the end user ?

nrobinaubertin: nvm, I understand that the user can give its own OCI images

sergiusens:
Hi, I thought I'd chime in. I was also pinged for it so here it goes :-)

A couple of things to make comment of and perhaps address:

- The `/bin`/`/usr/bin` issue is a bug, temporary and we prototyped the fix for it last week; it's on @tigarmo's list of things to do in the nearest future
- this looks a lot like the extension I proposed, is the intention to provide a generic image with only flask? Is that a good idea considering the amount of combinations of what makes flask a thing?
- My intention with the extension, even if a demoed using an Ubuntu base is to go the distroless route and use a bare base (I need to still explore any side effects of this path, especially with database connectors)
- We can probably clean up the left over venv files from the plugin before staging.

gregory-schiano:
> Hi, I thought I'd chime in. I was also pinged for it so here it goes :-)
> 
> A couple of things to make comment of and perhaps address:
> 
> * The `/bin`/`/usr/bin` issue is a bug, temporary and we prototyped the fix for it last week; it's on @tigarmo's list of things to do in the nearest future
> * this looks a lot like the extension I proposed, is the intention to provide a generic image with only flask? Is that a good idea considering the amount of combinations of what makes flask a thing?
> * My intention with the extension, even if a demoed using an Ubuntu base is to go the distroless route and use a bare base (I need to still explore any side effects of this path, especially with database connectors)
> * We can probably clean up the left over venv files from the plugin before staging.

The goal of this repo is to first build a "best-in-class" flask charm with the 12-factor in mind.
We want to cover it with integration tests, therefore we need to have an oci image available, hence the rockcraft.yaml
This rockcraft.yaml is also a way for us to find and experiment the rockcraft declaration to later feed the rockcraft flask-extension.
I'm fine with the extensions being at first based on Ubuntu, but I agree that we'll have to make it as light as possible, hence slices and bare

src/webserver.py
nrobinaubertin: I fixed that and added our main discussion points to the description of the PR

sample_rock/rockcraft.yaml
nrobinaubertin: See description of the PR on how we decided to proceed.

sample_rock/rockcraft.yaml
```diff
@@ -20,6 +20,11 @@ parts:
       - gunicorn
     python-requirements:
       - requirements.txt
+    override-prime: |
+      craftctl default
+      # We remove the standard executable to have only /bin/python3 in our PATH
+      # Related to https://github.com/canonical/rockcraft/issues/242
+      rm -f "$CRAFT_PRIME/usr/bin/python3"
```
gregory-schiano, line 27: I like it like that 👍

jdkandersson (approved):

weiiwang01 (approved):

PR #10, opened 2023-05-11
Add flask_* charm configurations for Flask builtin configurations

weiiwang01:


.licenserc.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - '**/.jujuignore'
+    - '**/.gitignore'
+    - '**/.licenserc.yaml'
+    - '**/.trivyignore'
+    - '**/CODEOWNERS'
+    - '**/icon.svg'
+    - '**/LICENSE'
+    - '**/trivy.yaml'
+    - 'lib/charms/traefik_k8s/v1/ingress.py'
+  comment: on-failure
```
yanksyoon: I think a new line is missing :D

src/charm.py
```diff
@@ -5,50 +5,96 @@
...
     def __init__(self, *args: typing.Any) -> None:
         """Initialize the instance.
 
         Args:
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
-        self.framework.observe(self.on.config_changed, self.config_service)
+        try:
+            self._charm_state = CharmState.from_charm(charm=self)
```
yanksyoon, line 36: I'm wondering if it would be better to explicitly pass on specific relation / databags, rather than passing the whole charm :)

jdkandersson:
I think this includes the changes from #2 ? It will be toufgh to review until that PR is merged

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+    description: >
+      Set the secure attribute in the Flask application cookies. This configuration will set the 
+      FLASK_SESSION_COOKIE_SECURE environment variable. You need to run `app.config.from_prefixed_env()` in your Flask
+      application in order to receive this configuration.
+  flask_preferred_url_scheme:
+    type: string
+    description: >
+      Use this scheme for generating external URLs when not in a request context in the Flask application. By default,
+      it's "HTTP" This configuration will set the  FLASK_PREFERRED_URL_SCHEME environment variable. You need to 
+      run `app.config.from_prefixed_env()` in your Flask application in order to receive this configuration.
```
yanksyoon: I think a new line is missing :D

src/charm.py
```diff
@@ -5,52 +5,108 @@
...
+        except CharmConfigInvalidError as exc:
+            self._update_status(BlockedStatus(exc.msg))
+            return
```
yanksyoon: I'm not quite sure if the state should raise error during init, would it be better if we had validation in the event handler?

src/charm.py
```diff
@@ -5,52 +5,108 @@
...
         Args:
             event: the config-changed event that trigger this callback function.
         """
-        container = self.unit.get_container("flask-app")
-        if not container.can_connect():
+        if not self.container_can_connect():
             event.defer()
             return
+
+        container = self.container()
```
yanksyoon: Just a nitpick, it feels a bit weird that self.container() already checks if it is connectable but we do a pre-check before.

src/charm_state.py
```diff
@@ -0,0 +1,201 @@
...
+# See LICENSE file for licensing details.
+
+"""This module defines the CharmState class which represents the state of the Flask charm."""
+
+import datetime
+import itertools
+import pathlib
+
+from ops.charm import CharmBase
+from pydantic import (  # pylint: disable=no-name-in-module
```
yanksyoon, line 13: Out of curiosity, what is causing no-name-in-module and I'm wondering if commenting the reason would help? :)

src/charm_state.py
```diff
@@ -0,0 +1,201 @@
...
+            keepalive=datetime.timedelta(seconds=int(self._webserver_keepalive))
+            if self._webserver_keepalive is not None
+            else None,
+            timeout=datetime.timedelta(seconds=int(self._webserver_timeout))
+            if self._webserver_timeout is not None
+            else None,
+        )
+
+    @property
+    def flask_config(self) -> dict[str, str | int | bool]:
```
yanksyoon, line 160: I'm wondering if typed dict would be a helpful return type to see what the necessary configurations are :D

src/charm_state.py
```diff
@@ -0,0 +1,201 @@
...
+        """Get the value of the flask_config charm configuration.
+
+        Returns:
+            The value of the flask_config charm configuration.
+        """
+        return self._flask_config.copy()
+
+    @property
+    def base_dir(self) -> pathlib.Path:
```
yanksyoon: Would these be better as constants since they are not quite "states"?

src/webserver.py
```diff
@@ -0,0 +1,156 @@
...
+
+        Args:
+            charm_state: The state of the charm that the GunicornWebserver instance belongs to.
+            flask_container: The Flask container in this charm unit.
+        """
+        self._charm_state = charm_state
+        self._flask_container = flask_container
+
+    @property
+    def _config(self) -> str:
```
yanksyoon: I'm wondering if jinja templating could be used here :)

yanksyoon (commented):
It looks great! Minor suggestions :D Thanks

config.yaml
```diff
@@ -1,3 +1,52 @@
...
 # See LICENSE file for licensing details.
-options: {}
+options:
+  webserver_workers:
+    type: int
+    description: The number of webserver worker processes for handling requests.
+  webserver_threads:
+    type: int
+    description: >
+      The number of worker threads for handling requests. Run each webserver worker with the specified number of threads.
```
arturo-seijas: Perhaps this is redundant? "Run each webserver worker with the specified number of threads."

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+  webserver_keepalive:
+    type: int
+    description: The number of seconds for webserver to wait for requests on a Keep-Alive connection.
```
arturo-seijas: s/The number of seconds for webserver to/Time in seconds to?

arturo-seijas: I'm thinking here about the resulting docs, this seems like simple grammar to me

gregory-schiano:
> I think this includes the changes from #2 ? It will be toufgh to review until that PR is merged

Yeap agree, better to focus on merging #2 and then review that one

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+  webserver_timeout:
+    type: int
+    description: Webserver workers silent for more than this many seconds are killed and restarted.
```
arturo-seijas: s/Webserver workers silent for more than this many seconds are killed and restarted./Time in seconds to kill silent workers?

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+  flask_env:
+    type: string
+    description: What environment the Flask app is running in, by default it's 'production'.
+  flask_debug:
+    type: boolean
+    description: Whether Flask debug mode is enabled.
+  flask_secret_key:
+    type: string
+    description: >
+      A secret key that will be used for securely signing the session cookie and can be used for any other security
```
arturo-seijas: s/that will be used/used
s/and can be used for any other security/ and user for other security

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+      related needs by your Flask application. This configuration will set the FLASK_SECRET_KEY environment variable.
+      You need to run `app.config.from_prefixed_env()` in your Flask application in order to receive this configuration.
+  flask_permanent_session_lifetime:
+    type: int
+    description: >
+      Set the cookie’s expiration to this number of seconds in the Flask application permanent sessions.
```
arturo-seijas: s/Set the cookie’s expiration to this number of seconds(Time in seconds for the cookie to expire?

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+    description: What environment the Flask app is running in, by default it's 'production'.
+  flask_debug:
+    type: boolean
+    description: Whether Flask debug mode is enabled.
+  flask_secret_key:
+    type: string
+    description: >
+      A secret key that will be used for securely signing the session cookie and can be used for any other security
+      related needs by your Flask application. This configuration will set the FLASK_SECRET_KEY environment variable.
+      You need to run `app.config.from_prefixed_env()` in your Flask application in order to receive this configuration.
```
arturo-seijas: s/You need to run/Run? Same for the rest of occurrences

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+  flask_permanent_session_lifetime:
+    type: int
+    description: >
+      Set the cookie’s expiration to this number of seconds in the Flask application permanent sessions. 
+      This configuration will set the FLASK_PERMANENT_SESSION_LIFETIME environment variable. You need to run 
+      `app.config.from_prefixed_env()` in your Flask application in order to receive this configuration.
+  flask_application_root:
+    type: string
+    description: >
+      Inform the Flask application what path it is mounted under by the application / web server. This configuration
```
arturo-seijas: s/Inform the Flask application what path it is mounted under by the application / web server/Path in which the application / web server is mounted?

config.yaml
```diff
@@ -1,3 +1,52 @@
...
+  flask_session_cookie_secure:
+    type: boolean
+    description: >
+      Set the secure attribute in the Flask application cookies. This configuration will set the 
+      FLASK_SESSION_COOKIE_SECURE environment variable. You need to run `app.config.from_prefixed_env()` in your Flask
+      application in order to receive this configuration.
+  flask_preferred_url_scheme:
+    type: string
+    description: >
+      Use this scheme for generating external URLs when not in a request context in the Flask application. By default,
```
arturo-seijas: s/Use this scheme/Scheme

src/charm_state.py
```diff
@@ -0,0 +1,201 @@
...
+        return "app:app"
+
+    @property
+    def flask_port(self) -> int:
+        """Gets the port number to use for the Flask server.
+
+        Returns:
+            The port number to use for the Flask server.
+        """
+        return 8000
```
arturo-seijas: This is already defined in a constant

src/consts.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""This module defines constants used throughout the Flask application."""
+
+FLASK_CONTAINER_NAME = "flask-app"
```
arturo-seijas, line 6: IMHO `FLASK_CONTAINER_NAME` should be defined in charm.py instead for better cohesion

src/consts.py
```diff
@@ -0,0 +1,8 @@
...
+FLASK_APP_PORT = 8000
+FLASK_SERVICE_NAME = "flask-app"
```
arturo-seijas, line 7: IMHO `FLASK_SERVICE_NAME` should be defined in charm.py instead for better cohesion

src/consts.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""This module defines constants used throughout the Flask application."""
+
+FLASK_CONTAINER_NAME = "flask-app"
+FLASK_APP_PORT = 8000
```
arturo-seijas: IMHO `FLASK_APP_PORT` should be defined in CharmState instead for better cohesion, and perhaps it should always be retrieved using the `flask_port` property

.licenserc.yaml
weiiwang01: Added, thanks!

src/charm.py
gregory-schiano: The method is meant to pass the whole charm, and is a kind of factory helper that'll get everything it needs from the charm, that's why we named it `from_charm`

src/charm.py
```diff
@@ -47,6 +53,16 @@ def __init__(self, *args: typing.Any) -> None:
...
 
+    def _update_status(self, status: StatusBase) -> None:
+        """Update the application and unit status.
+
+        Args:
+            status: the desired application and unit status.
+        """
+        self.unit.status = status
+        if self.unit.is_leader():
+            self.app.status = status
```
gregory-schiano, line 64: I'm not sure the app status should be the status of the leader... but perhaps an aggregation of the status of every unit... Didn't really think about it

src/charm_state.py
```diff
@@ -4,18 +4,67 @@
...
+    Field,
+    ValidationError,
+    validator,
+)
 
 from charm_types import WebserverConfig
+from exceptions import CharmConfigInvalidError
+
+
+class FlaskConfig(BaseModel, extra=Extra.allow):  # pylint: disable=too-few-public-methods
```
gregory-schiano, line 25: Can't we use it as a dataclass ?

src/charm_state.py
```diff
@@ -55,12 +108,27 @@ def from_charm(cls, charm: CharmBase) -> "CharmState":
...
+        try:
+            valid_flask_config = FlaskConfig(**flask_config)  # type: ignore
+        except ValidationError as exc:
+            error_fields = set(
+                itertools.chain.from_iterable(error["loc"] for error in exc.errors())
+            )
+            error_field_str = " ".join(f"flask_{f}" for f in error_fields)
+            raise CharmConfigInvalidError(f"invalid configuration: {error_field_str}") from exc
         return cls(
+            flask_config=valid_flask_config.dict(exclude_unset=True, exclude_none=True),
```
gregory-schiano, line 134: Any specific reason to convert it to a dict ? If FlaskConfig was a dataclass we could use it as is

src/charm_state.py
gregory-schiano: I agree, see my comments concerning dataclass

src/charm_state.py
```diff
@@ -4,18 +4,67 @@
...
+        application_root: inform the Flask application what path it is mounted under by the
+            application / web server.
+        session_cookie_secure: set the secure attribute in the Flask application cookies.
+        preferred_url_scheme: use this scheme for generating external URLs when not in a request
+            context in the Flask application.
+    """
+
+    env: str = Field(None, min_length=1)
+    debug: bool = Field(None)
+    secret_key: str = Field(None, min_length=1)
```
gregory-schiano: I think that if no secret_key is specified we have to generate one as Flask will block Session if that's not set

tests/unit/test_charm.py
```diff
@@ -28,3 +32,55 @@ def test_flask_pebble_layer(harness: Harness) -> None:
...
+) -> None:
+    """
+    arrange: none
+    act: set flask_* charm configurations.
+    assert: flask_config in the charm state should reflect changes in charm configurations.
+    """
+    harness.begin()
+    harness.update_config(charm_config)
+    charm_state = CharmState.from_charm(harness.charm)
+    assert charm_state.flask_config == flask_config
```
gregory-schiano: If we unittest the charmstate class I think this test is redundant
It's also covered by the integration test.
The only case we could test here is getting the pebble plan and ensure that the environment variables matches, but that's still redundant with the integration test

tests/unit/test_flask_app.py
```diff
@@ -0,0 +1,31 @@
...
+def test_flask_env(flask_config: dict):
+    """
+    arrange: create the Flask app object with a controlled charm state.
+    act: none.
+    assert: flask_environment generated by the Flask app object should be acceptable by Flask app.
+    """
+    charm_state = CharmState(flask_config=flask_config)
+    flask_app = FlaskApp(charm_state=charm_state)
+    assert flask_app.flask_environment == {
+        f"FLASK_{k.upper()}": json.dumps(v) for k, v in flask_config.items()
```
gregory-schiano: I think the `FLASK_` prefix could be a constant

gregory-schiano (commented):
I think unitests for the CharmState class are missing

config.yaml
weiiwang01: Added, thanks!

src/charm.py
weiiwang01: The reason I raise an error during the initialization is that I don't want the state to contain invalid data.

src/charm.py
weiiwang01: I updated the code to let the `container` method raises a `PebbleNotReadyError` and use try-catch for pebble availability issue. Thanks!

src/charm_state.py
weiiwang01: It's cause by pydantic and it's out of our control, I think the placement of this disable command tells that.

src/charm_state.py
weiiwang01: One advantage of properties is that it's immutable and it can be define as `abstractmethod` which I am considering doing it in the future. I think it's okay to leave this as is for now.

src/webserver.py
weiiwang01: This problem has been resolved in the last PR, thanks!

config.yaml

config.yaml

src/charm_state.py
weiiwang01: This is caused by merging, I will remove the constant, thanks!

src/consts.py
weiiwang01: Moved, thanks!

src/consts.py
weiiwang01: Moved, thanks!

src/consts.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: This is just a helper function to set both the unit status and app status to the same thing and it's not suited for other situation. I updated the function name to `_update_app_and_unit_status` to avoid confusions. Thanks!

src/charm_state.py
weiiwang01: I believe that charm should be consistent with Flask. If the charm does not have a secret key set, the charm managed flask should also not have a secret key set.

src/consts.py
weiiwang01: `FLASK_SERVICE_NAME` and `FLASK_CONTAINER_NAME` can't be move to `charm.py` since it will cause a circular import

src/consts.py
weiiwang01: `FLASK_SERVICE_NAME` and `FLASK_CONTAINER_NAME` can't be move to `charm.py` since it will cause a circular import

config.yaml

config.yaml

config.yaml

config.yaml

config.yaml

config.yaml

src/charm_state.py
weiiwang01: The problem is that after validation, there's no place in the code that access each individual fields inside `flask_config`. That's why I choose to use dict instead of data class to reduce the unnecessary complexity of the interface.

src/charm_state.py
weiiwang01: Answered in above.

tests/unit/test_flask_app.py
weiiwang01: I moved this to `consts.py`, thanks!

tests/unit/test_charm.py
weiiwang01: Answering one of the comment about missing unit tests for charm state.
I think this is one of the unit tests for charm state, I will move this to `test_charm_state.py`

src/consts.py
weiiwang01: FLASK_SERVICE_NAME and FLASK_CONTAINER_NAME can't be move to charm.py since it will cause a circular import

config.yaml
```diff
@@ -6,11 +6,46 @@ options:
...
     type: int
-    description: The number of seconds for webserver to wait for requests on a Keep-Alive connection.
+    description: Time in seconds for webserver to wait for requests on a Keep-Alive connection.
   webserver_timeout:
     type: int
-    description: Webserver workers silent for more than this many seconds are killed and restarted.
+    description: Time in seconds to kill and restart silent webserver workers.
+  flask_env:
+    type: string
+    description: What environment the Flask app is running in, by default it's 'production'.
```
jdkandersson, line 18: Would it be worth adding a `default` with `production` here?

config.yaml
```diff
@@ -6,11 +6,46 @@ options:
...
+  flask_debug:
+    type: boolean
+    description: Whether Flask debug mode is enabled.
+  flask_secret_key:
+    type: string
+    description: >
+      A secret key used for securely signing the session cookie and  for any other security related needs by your
```
jdkandersson: Are there multiple spaces between `and` and `for`?

config.yaml
```diff
@@ -6,11 +6,46 @@ options:
...
+    type: boolean
+    description: >
+      Set the secure attribute in the Flask application cookies. This configuration will set the 
+      FLASK_SESSION_COOKIE_SECURE environment variable. Run `app.config.from_prefixed_env()` in your Flask application
+      in order to receive this configuration.
+  flask_preferred_url_scheme:
+    type: string
+    description: >
+      Scheme for generating external URLs when not in a request context in the Flask application. By default,
+      it's "HTTP" This configuration will set the FLASK_PREFERRED_URL_SCHEME environment variable.
```
jdkandersson: Should this default to HTTPS?

requirements.txt
```diff
@@ -1 +1,2 @@
 ops >= 1.5.0
+pydantic
```
jdkandersson: Should we update the version pin for `ops` to be `2` and add at least a major version pin for `pydantic`?

src/charm_state.py
```diff
@@ -4,18 +4,67 @@
 """This module defines the CharmState class which represents the state of the Flask charm."""
 
 import datetime
+import itertools
 import pathlib
 
 from ops.charm import CharmBase
+from pydantic import (  # pylint: disable=no-name-in-module
```
jdkandersson, line 13: If `pydantic` is added to the dependencies list for the linting stage this should resolve this error, or it can just list the requirements file

src/charm_state.py
```diff
@@ -4,18 +4,67 @@
...
+        permanent_session_lifetime: set the cookie’s expiration to this number of seconds in the
+            Flask application permanent sessions.
+        application_root: inform the Flask application what path it is mounted under by the
+            application / web server.
+        session_cookie_secure: set the secure attribute in the Flask application cookies.
+        preferred_url_scheme: use this scheme for generating external URLs when not in a request
+            context in the Flask application.
+    """
+
+    env: str = Field(None, min_length=1)
```
jdkandersson: I'm not that familiar with `pydantic`, this reads like `env` could be `None` whereas the type hint indicates it is a `str`?

src/charm_state.py
```diff
@@ -24,6 +73,8 @@ class CharmState:
 
     def __init__(
         self,
+        *,
+        flask_config: dict[str, int | str] | None = None,
```
jdkandersson, line 79: What about using a `TypedDict` so that it is clear which keys are available?

src/charm_state.py
```diff
@@ -55,12 +108,27 @@ def from_charm(cls, charm: CharmBase) -> "CharmState":
...
         """
         keepalive = charm.config.get("webserver_keepalive")
         timeout = charm.config.get("webserver_timeout")
         workers = charm.config.get("webserver_workers")
         threads = charm.config.get("webserver_threads")
+        flask_config = {
+            k.removeprefix("flask_"): v for k, v in charm.config.items() if k.startswith("flask_")
+        }
+        try:
+            valid_flask_config = FlaskConfig(**flask_config)  # type: ignore
```
jdkandersson, line 126: What is the typing problem here please?

src/flask_app.py
```diff
@@ -0,0 +1,36 @@
...
+# See LICENSE file for licensing details.
+
+"""Provide the FlaskApp class to represent the Flask application."""
+import json
+
+from charm_state import CharmState
+from consts import FLASK_ENV_CONFIG_PREFIX
+
+
+class FlaskApp:  # pylint: disable=too-few-public-methods
```
jdkandersson, line 11: For this, and any other disable, should we add a comment explaining why the disable is required in this case?

tests/integration/test_charm.py
```diff
@@ -64,6 +65,68 @@ async def test_flask_webserver_timeout(
...
+        ),
+    ],
+    indirect=["update_config"],
+)
+@pytest.mark.usefixtures("update_config")
+async def test_invalid_flask_config(flask_app: Application, invalid_configs):
+    """
+    arrange: build and deploy the flask charm, and change flask related configurations
+        to certain invalid values.
+    act: none.
```
jdkandersson, line 117: Would the `act` be the change of configuration to an invalid value?

config.yaml
weiiwang01: The default production is implemented by Flask when the ENV configuration is not specified, so I think the charm should allow Flask to do that.

config.yaml

config.yaml
weiiwang01: The default value is decided by Flask.

src/charm_state.py
weiiwang01: I tried to add pydantic to the linting dependencies but it still displays an error.

src/charm_state.py
weiiwang01: There's no need to access each individual fields inside the `flask_config`, so a simple dict is sufficient.

src/charm_state.py
weiiwang01: Fixed, thanks!

src/charm_state.py
weiiwang01: pydantic model's constructor is typed with each individual field as keyword arguments, which is not compatible with dict unpacking.

src/flask_app.py
weiiwang01: I think the name `too-few-public-methods` is enough self-explanatory here?

tests/integration/test_charm.py
weiiwang01: Configuration change happens in the fixture, so I put it in the arrange section.

src/charm_state.py
weiiwang01: I added a comment for that, thanks!

config.yaml
jdkandersson: Should we be more opinionated and make the default HTTPS?

src/charm_state.py
jdkandersson: I misread the error, perhaps add a note here instead on why the disable is appropriate?

requirements.txt
weiiwang01: Updated, thanks!

src/charm_state.py
weiiwang01: Updated, thanks!

config.yaml
weiiwang01: I think it's best not to altering flask configuration defaults to prevent user confusion.

jdkandersson: @gregory-schiano? Any thoughts? What are the impacts of changing to HTTPS, does that mean the dev server doesn't work? One of the problems with security is unsafe defaults, and this is potentially one

arturo-seijas (dismissed):

src/charm_state.py
gregory-schiano: I think the charm should (must ?) be a bit more opinionated, and the secret_key being one. We should help the users to operate in the best way their application, and generating a best-practice secret_key if it's not provided is part of how to best operate a flask application

config.yaml
gregory-schiano: Similarly to "secret_key" I think we should be indeed opinionated here, and "HTTPS" is part of the best-practices of how to operate.

weiiwang01: I can change the default of FLASK_PREFERRED_URL_SCHEME to be HTTPS very easily. However, for flask_secret_key it requires a lot of efforts, since the secret key requires a persistent storage (peer relation) and we need to provide an action to reset secret key if there's a data breach. Considering the effort involved, we should evaluate whether the benefits outweigh the costs

Also, If we decide to implement charm generated secret, that will be out of the scope of this PR, we need to create another story for that.

src/charm_state.py
jdkandersson: Proposal is to implement this in a follow up story, could you create that story and we can make it out of scope for this pr?

config.yaml
jdkandersson: Please go ahead with implementing the HTTPS part, for the secret key I agree we should do it in a follow up story

weiiwang01: The default value of preferred_url_scheme has been changed to HTTPS, thanks!

jdkandersson (approved):

src/charm.py
```diff
@@ -11,11 +11,12 @@
 from charms.traefik_k8s.v1.ingress import IngressPerAppRequirer
 from ops.charm import CharmBase, ConfigChangedEvent
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, Container
+from ops.model import ActiveStatus, BlockedStatus, Container, StatusBase
 
 from charm_state import CharmState
-from consts import FLASK_APP_PORT, FLASK_CONTAINER_NAME, FLASK_SERVICE_NAME
-from exceptions import WebserverConfigInvalidError
+from consts import FLASK_CONTAINER_NAME, FLASK_SERVICE_NAME
```
merkata, line 17: Very nitpicky, though I should ask - constants is being abbreviated to consts, whereas exceptions remains (and contants is 9 chars, exceptions 10) - can the naming be consistent like either abbreviate all that can be abbreviated or abbreviate none of them?

src/charm.py
```diff
@@ -31,22 +32,37 @@ def __init__(self, *args: typing.Any) -> None:
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
-        self._charm_state = CharmState.from_charm(charm=self)
+        try:
+            self._charm_state = CharmState.from_charm(charm=self)
+        except CharmConfigInvalidError as exc:
```
merkata, line 37: Won't a linter complain on the variable name here, would something like `config_error` be better?

src/charm_state.py
gregory-schiano: Ok fine, I made a story

src/charm.py
gregory-schiano: Well it doesn't simply set the app and unit status.
I sets the unit status and if it's the leader sets the app status to same status of the leader unit

So if I get it correctly if a unit is failing but the leader is ok, the app status will be Active even if a unit has a problem.
I think we should have something a bit smarter for the app status

src/charm_state.py
gregory-schiano: Ok let's keep it like this and re-evaluate if we make some changes

src/charm.py
jdkandersson: I think `exc` is en exception to that 😉  it also lets things like `k` and `v` through

config.yaml
```diff
@@ -6,11 +6,47 @@ options:
...
+    description: >
+      Set the secure attribute in the Flask application cookies. This configuration will set the 
+      FLASK_SESSION_COOKIE_SECURE environment variable. Run `app.config.from_prefixed_env()` in your Flask application
+      in order to receive this configuration.
+  flask_preferred_url_scheme:
+    type: string
+    default: HTTPS
+    description: >-
+      Scheme for generating external URLs when not in a request context in the Flask application. By default,
+      it's "HTTPS" This configuration will set the FLASK_PREFERRED_URL_SCHEME environment variable.
```
nrobinaubertin, line 51: dot missing after "HTTPS"

nrobinaubertin (approved):

src/charm.py
weiiwang01: Gotcha. I will update the module name `consts` to `constants` in the next PR https://github.com/canonical/flask-k8s-operator/pulls since this PR is merging.

config.yaml
weiiwang01: I will update this in the next following PR https://github.com/canonical/flask-k8s-operator/pulls, thanks!

src/charm.py
weiiwang01: Yes, this is more of a Python convention.

src/charm.py
weiiwang01: The problem is that there's not yet a common synchronization method between units in juju. Even if the leader unit want to evaluate all unit status before setting the application status, there's no easy way to do that.

For now, all unit runs the same code and uses same OCI images, so in most cases, we can assume the states should be the same across all units. And, in a high availability deployment, if at least one copy is healthy, the entire application can consider as (downgraded) health, so in theory, we can set the application to be active if at least the leader is active.

I agree we should evaluate this problem in the future.

src/charm.py
merkata: Nice, TIL - thanks!

PR #12, opened 2023-05-15
Enable user-defined Flask application configs in the charm

weiiwang01:
With this update, the charm gains the ability to accept user-defined Flask app configurations. 
These configurations are injected into the charm using charmcraft during build time. However this feature is not yet implemented in charmcraft.


config.yaml
```diff
@@ -6,11 +6,46 @@ options:
...
+  flask_env:
+    type: string
+    description: What environment the Flask app is running in, by default it's 'production'.
+  flask_debug:
+    type: boolean
+    description: Whether Flask debug mode is enabled.
+  flask_secret_key:
+    type: string
+    description: >-
+      A secret key used for securely signing the session cookie and  for any other security related needs by your
```
yanksyoon: s/A secret/The secret

src/charm_state.py
```diff
@@ -85,6 +167,24 @@ def webserver_config(self) -> WebserverConfig:
...
         )
 
+    @property
+    def flask_config(self) -> dict[str, str | int | bool]:
+        """Get the value of the flask_config charm configuration.
+
+        Returns:
+            The value of the flask_config charm configuration.
+        """
+        return self._flask_config.copy()
```
yanksyoon, line 188: A minor question, is this to make sure that mutation doesn't happen?

tests/integration/test_charm.py
```diff
@@ -64,6 +65,92 @@ async def test_flask_webserver_timeout(
...
+            {"flask_permanent_session_lifetime": 100},
+            {"PERMANENT_SESSION_LIFETIME": 100},
+            id="permanent_session_lifetime",
+        ),
+        pytest.param({"flask_debug": True}, {"DEBUG": True}, id="debug"),
+    ],
+    indirect=["update_config"],
+)
+@pytest.mark.usefixtures("update_config")
+async def test_flask_config(flask_app, get_unit_ips, excepted_config):
```
yanksyoon: Would the arguments be better if type hinted? :) same for other tests

yanksyoon (dismissed):
Minor suggestions, thanks!

src/charm_state.py
weiiwang01: Yes, it's, since there's not frozen dict in Python.

config.yaml

tests/integration/test_charm.py
weiiwang01: Updated, thanks!

config.yaml
```diff
@@ -6,11 +6,46 @@ options:
...
   webserver_timeout:
     type: int
-    description: Webserver workers silent for more than this many seconds are killed and restarted.
+    description: Time in seconds to kill and restart silent webserver workers.
+  flask_env:
+    type: string
+    description: What environment the Flask app is running in, by default it's 'production'.
+  flask_debug:
+    type: boolean
+    description: Whether Flask debug mode is enabled.
```
arturo-seijas, line 13: s/Whether Flask debug mode is enabled/Flask debug mode enabled

config.yaml
```diff
@@ -6,11 +6,46 @@ options:
...
+  flask_secret_key:
+    type: string
+    description: >-
+      The secret key used for securely signing the session cookie and  for any other security related needs by your
```
arturo-seijas: There seems to be a double space in `and for`

src/charm.py
```diff
@@ -62,11 +78,11 @@ def container(self) -> Container:
             The controller of the flask application workload container.
 
         Raises:
-            RuntimeError: if the pebble service inside the container is not ready while the
+            PebbleNotReadyError: if the pebble service inside the container is not ready while the
                 ``require_connected`` is set to True.
         """
         if not self.container_can_connect():
-            raise RuntimeError("pebble inside flask-app container is not ready")
+            raise PebbleNotReadyError("pebble inside flask-app container is not ready")
```
arturo-seijas: Why not logging and deferring the event here instead of throwing an exception?

src/consts.py
```diff
@@ -4,5 +4,18 @@
 """This module defines constants used throughout the Flask application."""
 
 FLASK_CONTAINER_NAME = "flask-app"
-FLASK_APP_PORT = 8000
 FLASK_SERVICE_NAME = "flask-app"
+FLASK_ENV_CONFIG_PREFIX = "FLASK_"
```
arturo-seijas: IHMO `FLASK_ENV_CONFIG_PREFIX`and `KNOWN_CHARM_CONFIG` would be better defined in `flask_app.py` and `charm_state.py`, respectively, for higher cohesion

src/consts.py
```diff
@@ -4,5 +4,18 @@
 """This module defines constants used throughout the Flask application."""
 
 FLASK_CONTAINER_NAME = "flask-app"
```
arturo-seijas: `FLASK_CONTAINER_NAME` and `FLASK_SERVICE_NAME` could be defined in `charm.py` instead

src/exceptions.py
```diff
@@ -4,17 +4,21 @@
 """Exceptions used by the Flask charm."""
 
 
-class WebserverConfigInvalidError(Exception):
-    """Exception raised when a webserver configuration is found to be invalid.
+class CharmConfigInvalidError(Exception):
```
arturo-seijas: I'd define this exception in `charm_state.py` for higher cohesion

src/charm.py
weiiwang01: This method can be used in different context, and not all context have the access to the event object.
Also, charm should stop executing any code after the event deferring, which is not possible to achieve inside the `container` method.

config.yaml

src/consts.py
weiiwang01: Updated, thanks!

src/consts.py
weiiwang01: It can't be defined in `charm.py` since that will cause circular import.

src/exceptions.py
weiiwang01: The exceptions are used in `charm_state.py` and `webserver.py`

config.yaml
weiiwang01: I think I will keep the current version for it's a complete sentences.

arturo-seijas (dismissed):

src/charm_state.py
jdkandersson: Is a shallow copy enough here?

src/charm_state.py
```diff
@@ -202,3 +223,18 @@ def flask_port(self) -> int:
             The port number to use for the Flask server.
         """
         return 8000
+
+
+KNOWN_CHARM_CONFIG = (
```
jdkandersson: Do you think these should be read from `config.yaml`?

src/charm_state.py
weiiwang01: Yes, the `flask_config` is always a mapping from string to string, integer or boolean.

tests/integration/test_charm.py
```diff
@@ -144,7 +176,7 @@ async def test_with_ingress(
     # mypy doesn't see that ActiveStatus has a name
     await model.wait_for_idle(status=ActiveStatus.name)  # type: ignore
 
-    traefik_ip = next(await get_unit_ips(traefik_app_name))
+    traefik_ip = next(iter(await get_unit_ips(traefik_app_name)))
```
jdkandersson: Out of curiosity, what is the reason for needing the `iter` here?

src/charm_state.py
weiiwang01: Since the user-defined configurations are added to the `config.yaml` during the build time, reading from `config.yaml` during the runtime will include the user-defined configurations too.

tests/integration/test_charm.py
weiiwang01: It's due to that `get_unit_ips` returns a tuple, I think I will change this to `(await get_unit_ips(traefik_app_name)[0]`, thanks!

src/charm_state.py
jdkandersson: Do you have any ideas for how we could reduce the duplication of information? Or manage the risk of divergence?

weiiwang01: There's a unit test `test_known_charm_config` that will check the consistency between the `KNOW_CHARM_CONFIG` and `config.yaml`.

jdkandersson (dismissed):

merkata (dismissed):
Great!

src/charm_state.py
```diff
@@ -119,10 +124,16 @@ def from_charm(cls, charm: CharmBase) -> "CharmState":
...
         flask_config = {
-            k.removeprefix("flask_"): v for k, v in charm.config.items() if k.startswith("flask_")
+            k.removeprefix("flask_"): v
+            for k, v in charm.config.items()
+            if k.startswith("flask_") and k in KNOWN_CHARM_CONFIG
+        }
+        app_config = {
+            k.removeprefix("flask_"): v
+            for k, v in charm.config.items()
+            if k.startswith("flask_") and k not in KNOWN_CHARM_CONFIG
```
gregory-schiano: I think there's a problem here:
1) I don't think we should remove any prefix from the app configuration, if the user used "flask_" they'll endup with an envvar "FLASK_FLASK_", we should keen them as they've been defined
2) Filtering out "known_charm_config" is usefull there, but doesn't guarantee potential envvar clash. Let's say a user defined a configuration named "application_root", this will be translated to "FLASK_APPLICATION_ROOT" which gets in conflict with a known_charm_config. Therefore a further filtering is required in the app_config

src/charm_state.py
```diff
@@ -119,10 +124,16 @@ def from_charm(cls, charm: CharmBase) -> "CharmState":
...
         }
         try:
-            # typing issue of the pydantic model constructor
```
gregory-schiano, line 125: Why did you remove this comment, the "type: ignore" is still needed and there

src/charm_state.py
```diff
@@ -202,3 +223,18 @@ def flask_port(self) -> int:
             The port number to use for the Flask server.
         """
         return 8000
+
+
+KNOWN_CHARM_CONFIG = (
+    "webserver_workers",
+    "webserver_threads",
+    "webserver_keepalive",
+    "webserver_timeout",
```
gregory-schiano: I don't think we need to define these parameters here as they can't get in conflict

tests/integration/conftest.py
```diff
@@ -33,20 +36,56 @@ def traefik_app_name_fixture() -> str:
     return "traefik-k8s"
 
 
+@pytest_asyncio.fixture(scope="module", name="build_charm")
+async def build_charm_fixture(ops_test) -> str:
+    """Builds the charm and injects additional configurations into config.yaml.
+
+    This fixture is designed to simulate a feature that is not yet available in charmcraft that
+    allows for the modification of charm configurations during the build process.
+    Three additional configurations, namely flask_foo_str, flask_foo_int, and flask_foo_bool,
```
gregory-schiano: The additional configuration aren't necessarily prefixed with "flask_" so I'd suggest to remove the prefix and keep "foo_str", "foo_int" and "foo_bool"
I'd also add a variable that will get in conflict with existing flask configuration like "application_root" to ensure the "flask_application_root" config doesn't get overriden

tests/integration/conftest.py
```diff
@@ -33,20 +36,56 @@ def traefik_app_name_fixture() -> str:
...
+    new_charm = io.BytesIO()
+    with zipfile.ZipFile(new_charm, "w") as new_charm_zip:
+        for item in charm_zip.infolist():
+            if item.filename == "config.yaml":
+                new_charm_zip.writestr(item, modified_config)
+            else:
+                with charm_zip.open(item) as file:
+                    data = file.read()
+                new_charm_zip.writestr(item, data)
+    charm_zip.close()
```
gregory-schiano, line 71: Isn't there a way to replace only `config.yaml` instead of recreating a new zip ?

tests/integration/test_charm.py
```diff
@@ -127,13 +131,41 @@ async def test_invalid_flask_config(flask_app: Application, invalid_configs):
             assert invalid_config in unit.workload_status_message
 
 
+@pytest.mark.parametrize(
+    "update_config, excepted_config",
+    [
+        pytest.param({"flask_foo_str": "testing"}, {"FOO_STR": "testing"}, id="str"),
+        pytest.param({"flask_foo_int": 128}, {"FOO_INT": 128}, id="int"),
+        pytest.param({"flask_foo_bool": True}, {"FOO_BOOL": True}, id="bool"),
```
gregory-schiano: As specified in a previous comment, we shouldn't remove the "flask_" prefix from the configuration, if a user specified "flask_foo_str" the envvar will be "FLASK_FLASK_FOO_STR"

tests/integration/conftest.py
gregory-schiano: We should probably also define a more complex configuration as even if not encouraged, some users may define an array or a dict as configuration parameters and we should make sure this is working.
Even if that's pretty similar to "str", I think we should cover this

tests/integration/test_charm.py
```diff
@@ -144,7 +176,7 @@ async def test_with_ingress(
     # mypy doesn't see that ActiveStatus has a name
     await model.wait_for_idle(status=ActiveStatus.name)  # type: ignore
 
-    traefik_ip = next(await get_unit_ips(traefik_app_name))
+    traefik_ip = (await get_unit_ips(traefik_app_name))[0]
```
gregory-schiano, line 182: Any reason for this change ?

src/charm_state.py
weiiwang01: 1. Should we strip the `flask_` prefix for other builtin charm configuration, like `flask_ application_root` as well? For the sake of consistency?
2. The configuration conflict and merging is handled in the `FlaskApp.flask_environment` function.

src/charm_state.py
weiiwang01: It's accidentally removed during merging, I will restore that, thanks!

src/charm_state.py
weiiwang01: This list is just used for filtering out user-defined configurations, the conflict handling happens in `FlaskApp.flask_environment`. The reason I didn't perform the merging and conflict removing earlier is to preserve as much information as possible for charm to decide what to do.

tests/integration/conftest.py
weiiwang01: Python `zipfile` library doesn't support modify zip file in place.

tests/integration/test_charm.py
weiiwang01: The new one is just simpler.

tests/integration/conftest.py
weiiwang01: After we switch to flatted config list, I am using the charm config type to determine user-defined configurations' types. But juju charm config type doesn't support dictionary values. Should we treat all user-defined configurations as JSON string? For example, if user want to set the `foo_str` user has to wrap the `foo_str` value inside double quotes. `juju config flask-k8s foo_str='"bar"'`

src/charm_state.py
gregory-schiano: For 1) I'm not sure I get your point, you're already removing the prefix in the flask_config params.
What I want to highlight here is that we don't expect app_config to start with "flask_" and we don't modify their name (so don't remove any potential prefix).
So if a user set a config:
`my_app_salt` that's accepted and translated to `FLASK_MY_APP_SALT` envvar
If a user set a config:
`flask_admin_email` that's accepted and translated to `FLASK_FLASK_ADMIN_EMAIL`

The builtin config are stripped to have a consistent environment transformation (add prefix + uppercase)

For 2) Indeed I saw it while doing the review

src/charm_state.py
gregory-schiano: What I meant is that a user won't be able to define a `webserver_workers` configuration as the end result will be 2 charm config with `webserver_workers` and in such case I expect charmcraft to raise the error.
So this won't reach the charm code, therefore this case will never happen

tests/integration/test_charm.py
gregory-schiano: Ok, that probably just me who's not fan of accessing arrays by index :)

src/charm_state.py
weiiwang01: I mean the configuration names `flask_application_root` defined in the `config.yaml` file, should we remove the `flask_` from the built-in configuration names defined in the `config.yaml` file as well?

tests/integration/conftest.py
gregory-schiano: I don't think we should treat all user-defined configuration as JSON string, but we should consider the scenario where a user will then as JSON string.
Therefore if a user defines a variable `foo_dict` and set it to `juju config flask-k8s foo_dict='{ "foo": "bar", "bar": "foo" }'` or same with `foo_array` using `juju config flask-k8s foo_array='["foo", "bar", "foobar"]'`
We should ensure that the charm is properly working and passing this value to the `FLASK_FOO_DICT` and `FLASK_FOO_ARRAY` envvar without breaking the envvar or the charm

src/charm_state.py
weiiwang01: Using a built-in configuration name to exclude all built-in configurations leaving the user-defined configuration is a simple and one-step solution. I understand `webserver_*` will not be possible to be defined by users, but there's no need to have any special treatment for `webserver_*` than other built-in configurations.

tests/integration/conftest.py
weiiwang01: What if the user want a configuration that is a string, but  `string == "{}"`?  How should differentiate that user want a string or dict in that case?

src/charm_state.py
gregory-schiano: I don't think so, the prefix in these configuration are there because they are specific to flask
The prefix for the envvar are there by convention for Flask to know which envvars to use

Even if the 2 prefixes are the same, they serve different purpose

weiiwang01:
Talked with @gregory-schiano and settled some design changes.

merkata:
> Talked with @gregory-schiano and settled some design changes.

Could you summarize in one or two sentences what these changes are? Thanks!

src/charm_state.py
```diff
@@ -21,6 +22,20 @@
 from charm_types import WebserverConfig
 from exceptions import CharmConfigInvalidError
 
+KNOWN_CHARM_CONFIG = (
```
merkata, line 25: It doesn't have to be, but the list of config variables is not sorted, perhaps it could add to readability.

src/charm_state.py
```diff
@@ -119,10 +138,12 @@ def from_charm(cls, charm: CharmBase) -> "CharmState":
         workers = charm.config.get("webserver_workers")
         threads = charm.config.get("webserver_threads")
         flask_config = {
-            k.removeprefix("flask_"): v for k, v in charm.config.items() if k.startswith("flask_")
+            k.removeprefix("flask_"): v
```
merkata, line 141: I'd document that behavior in the docstring on dealing with variables.

merkata (dismissed):

weiiwang01:
> > Talked with @gregory-schiano and settled some design changes.
> 
> Could you summarize in one or two sentences what these changes are? Thanks!

Sure, here are two changes:
1. the user-defined application configurations will drop the `flask_` prefix
2. when pass the string typed configurations to Flask, the configuration value will be passed as the face value instead of JSON encoded.

src/charm_state.py
weiiwang01: Sorted, and I sorted the configurations in `config.yaml` too.

src/charm_state.py
weiiwang01: Added, thanks!

merkata (approved):

jdkandersson (approved):

PR #13, opened 2023-05-18
Add Canonical Observability Stack integration

weiiwang01:


src/charm.py
```diff
@@ -87,6 +93,14 @@ def container(self) -> Container:
         container = self.unit.get_container(FLASK_CONTAINER_NAME)
         return container
 
+    def _prepare_flask_log_dir(self) -> None:
```
jdkandersson: Should this be in the flask module?

src/charm.py
```diff
@@ -127,6 +142,32 @@ def flask_layer(self) -> dict:
...
+        statsd_container = self.unit.get_container("statsd-prometheus-exporter")
+        statsd_layer = {
+            "summary": "statsd exporter layer",
+            "description": "statsd exporter layer",
+            "services": {
+                "statsd-prometheus-exporter": {
+                    "override": "replace",
+                    "summary": "statsd exporter service",
+                    "user": "nobody",
+                    "command": "/bin/statsd_exporter",
```
jdkandersson, line 147: Similar to the flask layer, should more of this come from an Observer module?

jdkandersson (commented):
Looking good

src/charm.py
weiiwang01: This is pebble layer, and this is a statsd prometheus exporter, which is not a part of canonical observability stack., so I think this should belong to the `charm.py`.

src/charm.py
weiiwang01: The flask module (`webserver.py`) receives the designated log path from charm, so I think it's the charm's duty to prepare the log path before passing the log path to webserver.

tests/integration/test_charm.py
```diff
@@ -3,21 +3,26 @@
...
 from pytest_operator.plugin import OpsTest
 
 # caused by pytest fixtures
 # pylint: disable=too-many-arguments
 
 
+# caused by pytest fixtures
+# pylint: disable=too-many-arguments
+
+
```
mthaddon: This seems to be a duplication of the section above?

weiiwang01: Yes, it must be caused by merging, removed, thanks!

src/charm.py
```diff
@@ -127,6 +142,32 @@ def flask_layer(self) -> dict:
...
+                    "startup": "enabled",
+                }
+            },
+            "checks": {
```
amandahla, line 151: Do you think that should go to the flask module (webserver)?
The same comment was made [here](https://github.com/canonical/github-actions-exporter-operator/pull/3#discussion_r1193309642).

src/observability.py
```diff
@@ -0,0 +1,37 @@
...
+        self._metrics_endpoint = MetricsEndpointProvider(
+            charm,
+            relation_name="metrics-endpoint",
+            jobs=[{"static_configs": [{"targets": ["*:80"]}]}],
+        )
+        self._logging = LogProxyConsumer(
+            charm,
+            relation_name="logging",
+            log_files=[str(charm_state.flask_access_log), str(charm_state.flask_error_log)],
+            container_name="flask-app",
```
amandahla: Is possible to use the FLASK_CONTAINER_NAME constant here?

src/observability.py
```diff
@@ -0,0 +1,37 @@
...
+        """Initialize a new instance of the Observability class.
+
+        Args:
+            charm: The charm object that the Observability instance belongs to.
+            charm_state: The state of the charm that the Observability instance belongs to.
+        """
+        self._metrics_endpoint = MetricsEndpointProvider(
+            charm,
+            relation_name="metrics-endpoint",
+            jobs=[{"static_configs": [{"targets": ["*:80"]}]}],
```
amandahla: Is 80 the statsd exporter port?

src/webserver.py
```diff
@@ -59,6 +59,9 @@ def _config(self) -> str:
         config = f"""\
 bind = ['0.0.0.0:{self._charm_state.flask_port}']
 chdir = {repr(str(self._charm_state.flask_dir.absolute()))}
+accesslog = {repr(str(self._charm_state.flask_access_log.absolute()))}
```
amandahla, line 62: nitpick: access_log/error_log?

src/charm.py
jdkandersson: I can see why there is a reason to go one way or the other, the way logging works seems to be more application specific than general. That is why it would make more sense for the webserver module to be responsible for setting up the logging directory and anything else the application needs to be able to log. The webserver module can then say to the charm and any other integrations "here is where you can find the logs for the workload". Any thoughts on that?

src/charm.py
jdkandersson: Statsd seems to be linked to observability even if it isn't directly part of the observability stack? I'm guessing it is running to extract stats to be forwarded to the observability stack? Or is it used for other things as well?

src/charm.py
weiiwang01: If I put the flask logging directory preparation inside the the webserver module, then the `Observability` class will need to accept a webserver object as argument, is that okay?

src/charm.py
weiiwang01: The observability stack accepts only prometheus for metrics, the statsd export is for converting statsd metrics to prometheus metrics since the Gunicorn only support statsd natively.

src/charm.py
jdkandersson: The webserver state? I think that is exactly the purpose of the state yes

src/charm.py
jdkandersson: Would it make sense to locate the configuration for statsd in the observer module since it seems to be closely related to observability?

src/charm.py
weiiwang01: Not the state, but the webserver object, since the logging directory is managed by webserver now. It will be something like this

```python
class Observability:  # pylint: disable=too-few-public-methods
    def __init__(self, charm: CharmBase, charm_state: CharmState, webserver: GunicornWebserver):
```

jdkandersson: Could we make the logging directory part of the state so that we just need to pass the webserver state object to `Obervability`?

src/charm.py
weiiwang01: There's no configuration for statsd, the only thing needed for starting statsd is the statsd pebble layer. Should we put the statsd pebble layer inside the observability module?

jdkandersson: It would be good to use constants from the observability module to generate the pebble layer, like make the command a constant in the observability module and use the constant to create the observability layer in `charm.py`. Does that make sense?

src/charm.py
weiiwang01: Currently, we are putting the logging directory inside the `CharmState`, but since the charm state is derived from the charm, the preparation of the logging directory (mkdir in workload container) is done by the charm.

src/charm.py
weiiwang01: The problem is that the statsd command to start the statsd service is not tied to the observability stack, instead it is based on this resource definition in the `metadata.yaml` file. For different webservers, they might use different exports other than statsd exporter. That's why I think this should be the charm's responsibility for decide the command of the exporter pebble layer instead of observability module.

```yaml
  statsd-prometheus-exporter-image:
    type: oci-image
    description: Prometheus exporter for statsd data
    auto-fetch: true
    upstream-source: prom/statsd-exporter
```

src/observability.py
weiiwang01: Updated, thanks!

src/observability.py
weiiwang01: No, I think I copied the wrong configuration, fixed, thanks!

src/webserver.py
weiiwang01: Those are configuration fields required by gunicorn, we can't change that unfortunately.

src/charm.py
weiiwang01: Please see the discussion here: https://github.com/canonical/flask-k8s-operator/pull/13#discussion_r1198574412
Thanks!

src/charm.py
weiiwang01: I moved the preparation of logging directories to the webserver module, thanks!

jdkandersson (approved):

merkata (approved):
Looks good!

PR #14, opened 2023-05-22
Integrate the postgresql and mysql databases

nrobinaubertin:
The goal is to allow integration with postgresql and mysql through their respective postgresql_client and mysql_client interfaces.


flask_test_app/app.py
jdkandersson, line 1: Isn't there an existing one of these?

jdkandersson: And if not, should this be defined within the `test` directory?

flask_test_app/app.py
```diff
@@ -0,0 +1,116 @@
...
+from flask import Flask, g, jsonify, request
+
+app = Flask(__name__)
+app.config.from_prefixed_env()
+
+
+def get_mysql_database():
+    """Get the mysql db connection."""
+    if "mysql_db" not in g:
+        if app.config.get("MYSQL_DB_CONNECT_STRING"):
```
jdkandersson: Shouldn't this be an environment variable?

flask_test_app/app.py
```diff
@@ -0,0 +1,116 @@
...
+            return None
+    return g.mysql_db
+
+
+def get_postgresql_database():
+    """Get the postgresql db connection."""
+    if "postgresql_db" not in g:
+        if app.config.get("POSTGRESQL_DB_CONNECT_STRING"):
+            uri_parts = urlparse(app.config.get("POSTGRESQL_DB_CONNECT_STRING"))
+            g.postgresql_db = psycopg.connect(
```
jdkandersson: Do these support passing in the connection string as well?

flask_test_app/app.py
```diff
@@ -0,0 +1,116 @@
...
+                password=uri_parts.password,
+                database=uri_parts.path[1:],
+                port=uri_parts.port,
+            )
+        else:
+            return None
+    return g.mysql_db
+
+
+def get_postgresql_database():
```
jdkandersson: Should this code comply with the contributing guide (e.g., type hints, docstrings etc)?

flask_test_app/app.py
```diff
@@ -0,0 +1,116 @@
...
+                dbname=uri_parts.path[1:],
+                port=uri_parts.port,
+            )
+        else:
+            return None
+    return g.postgresql_db
+
+
+@app.teardown_appcontext
+def teardown_database(_):
```
jdkandersson: Aren't there libraries that do all of this for you? Like https://flask-mysql.readthedocs.io/en/stable/

flask_test_app/rockcraft.yaml
```diff
@@ -16,6 +16,10 @@ parts:
     source: .
     stage-packages:
       - python3-venv
+      - bash
```
jdkandersson: Are all these needed?

src/charm.py
```diff
@@ -52,6 +53,32 @@ def __init__(self, *args: typing.Any) -> None:
             host=f"{self.app.name}-endpoints.{self.model.name}.svc.cluster.local",
             strip_prefix=True,
         )
+        self.databases: typing.Dict[str, DatabaseRequires] = {
```
jdkandersson: @yanksyoon came up with an observer pattern, would that be applicable here?

src/charm.py
```diff
@@ -109,6 +136,61 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         container.replan()
         self._update_app_and_unit_status(ActiveStatus())
 
+    def _database_uri(self) -> typing.Dict[str, str]:
```
jdkandersson: Should this be in a separate database module?

src/charm.py
```diff
@@ -109,6 +136,61 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+            if not relation_data:
+                continue
+
+            # There can be only one database integrated at a time
+            # see: metadata.yaml
+            data = relation_data[0]
+
+            # Check that the relation data is well formed according to the following json_schema:
+            # https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/mysql_client/v0/schemas/provider.json
+            if not all(data.get(key) for key in ("endpoints", "username", "password")):
```
jdkandersson: Should this be an exception that puts the charm in blocked state?

src/charm.py
```diff
@@ -109,6 +136,61 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
+            endpoint = data["endpoints"].split(",")[0]
+            db_uris[database] = (
+                f"{database}://"
+                f"{data['username']}:{data['password']}"
+                f"@{endpoint}/{database_name}"
+            )
+
+        return db_uris
+
+    def _get_flask_env_config(self) -> dict[str, str]:
```
jdkandersson: Should this be in the webserver module?

src/charm.py
```diff
@@ -121,8 +203,8 @@ def flask_layer(self) -> dict:
                     "override": "replace",
                     "summary": "Flask application service",
                     "command": shlex.join(self._webserver.command),
-                    "environment": self._flask_app.flask_environment,
                     "startup": "enabled",
+                    "environment": self._get_flask_env_config(),
```
jdkandersson: What is the reason for changing this please?

tests/integration/test_charm.py
```diff
@@ -187,3 +188,45 @@ async def test_with_ingress(
...
     assert "Hello, World!" in response.text
+
+
+@pytest.mark.parametrize(
+    "endpoint,db_name, db_channel",
+    [
+        (
+            "mysql/status",
+            "mysql-k8s",
+            "8.0/stable",
```
jdkandersson: If the trailing comma on this line is deleted `black` will probably reduce the number of lines this is spread over

tests/integration/test_charm.py
```diff
@@ -187,3 +188,45 @@ async def test_with_ingress(
...
+    flask_app: Application,
+    model: juju.model.Model,
+    get_unit_ips,
+    endpoint: str,
+    db_name: str,
+    db_channel: str,
+):
+    """
+    arrange: build and deploy the flask charm.
+    act: deploy the ingress, configure it and relate it to the charm.
```
jdkandersson: Is this accurate? Is this test for the database or ingress?

tox.ini
```diff
@@ -67,7 +67,8 @@ commands =
       --skip {toxinidir}/lib \
       --skip {toxinidir}/venv \
       --skip {toxinidir}/.mypy_cache \
-      --skip {toxinidir}/tests/integration/.mypy_cache
+      --skip {toxinidir}/tests/integration/.mypy_cache \
+      --skip {toxinidir}/tests/integration/flask_test_app/.mypy_cache
```
jdkandersson: What is the reason for skipping this please?

nrobinaubertin: This is a left-over of me trying to put it in the `test/integration` directory. See [this comment](https://github.com/canonical/flask-k8s-operator/pull/14#discussion_r1201581123)

jdkandersson: So I guess we can remove it?

flask_test_app/app.py
nrobinaubertin: I just renamed the existing one. I've tried to put it in the test/integration directory but mypy is complaining about missing types and [doesn't want to install them in the CI](https://github.com/canonical/flask-k8s-operator/actions/runs/5045847451/jobs/9050633587#step:4:18).  
So I decided to place the directory where it was before for simplicity

flask_test_app/app.py
nrobinaubertin: It hadn't before so I wasn't too focused on that. I can take some time for that

flask_test_app/app.py
jdkandersson: It would be good to figure out how to resolve that error

flask_test_app/app.py
jdkandersson: It is starting to be a bit more complex which is when type hints can start to become helpful, though that can be mitigated using a library like flask-mysql to reduce the complexity

flask_test_app/app.py
nrobinaubertin: I don't understand: this is an environment variable processed by flask ?

flask_test_app/app.py
nrobinaubertin: For our usage, I could even remove this function. I like having things explicit when they are so simple and in tests.

flask_test_app/app.py
jdkandersson: I suppose the question is whether this should be processed by flask or if it should be an environment variable that doers not get picked up by flask configuration. The reason is that the database connection string contains sensitive information and hence generally shouldn't be in configuration and instead be passed using just environment variables

jdkandersson: Similar to that sensitive values in a terraform file are passed using a vault rather than written into the terraform file

flask_test_app/app.py
jdkandersson: it isn't just this it is also the rest of the other database functions, most users would also use flask-mysql so a good side effect is to check what the experience would be with that

flask_test_app/app.py
nrobinaubertin: Fixed

flask_test_app/rockcraft.yaml
nrobinaubertin: Removed

tests/integration/test_charm.py
nrobinaubertin: Fixed

tox.ini
nrobinaubertin: Removed

tests/integration/test_charm.py
nrobinaubertin: Fixed

src/charm.py
nrobinaubertin: Can you point me to an accepted example of this pattern in our charms so that I use it as a basis for consideration ?

flask_test_app/app.py
nrobinaubertin: I think most would use sqlalchemy (7.2k stars) or pymysql (7.3k stars) instead of flask-mysql (116 stars). I choose to not use sqlalchemy to keep things simple

flask_test_app/app.py
nrobinaubertin: For now, I'm strictly complying to the spec. This should be decided at the spec level.

> The database connection string will be available in the POSTGRES_DB_CONNECT_STRING environment variable of the flask application

jdkandersson: From my interpretation that line means that it shouldn't end up in the flask configuration, it should just be an environment variable in the container and that was my intent when writing the spec

flask_test_app/app.py
jdkandersson: There is also https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/ which is the library that I should have linked to initially

src/charm.py
jdkandersson: Perhaps ask Yanks since he came up with the proposal 🙂

flask_test_app/app.py
nrobinaubertin: I still think that using a more wildly used library and access the database directly without implementing models is the best combo for these tests.

flask_test_app/app.py
nrobinaubertin: I'm not sure to understand how the flask application can configure its connection to a db without access to the db_uri. And the availability to flask of a env variable does not differ if it's processed by flask's config mechanism (`os.environ` is always available). What is the threat model we're are addressing by making it not available through flask's config ?

flask_test_app/app.py
merkata: That's a good point on figuring this out. It seems it already landed and renaming it made it resurface - good thing. Now my proposal is to track it in an issue and tackle it separately.

flask_test_app/app.py
merkata: Alright, good points, just trying to figure out whether I understand things correctly:

* looking at the code, my understanding is that flask is constructing a config from environment variables
* that config is not persisted

Now if these two are valid, the proposal is to not have flask know where to connect to via direct access to an environment variable, rather a middleware would get that info (from an environment variable) and would present it to flask, correct? I'm wondering what is the benefit here.
If there are security improvements, they should be noted - to my understanding you will always end up with accessing env variables and keeping them in volatile memory (again, correct me if I'm wrong here). If there are no security improvements, following flask spec might be a good thing - even if we give a PaaS solution, twisting the design of a Flask app from something expected (and documented) to something else should be well reasoned about and well documented.

flask_test_app/app.py
merkata: I think this is a discussion point that deserves some attention. Do we want to have type hints to help us maintain that test application or do we want to have type hints to say that we feel that with a certain degree of complexity an app needs to have type hints? With our code (like src/charm.py) the answer is clear, we type hint everything and that's good. For the test app, it might serve it's purpose without type hints as initially created. I'm not sure whether it will get to a really complex state, being a test app that is. Even `charmcraft init` will create you a (test) charm that doesn't have any type hints and leaves it to you, we might want to do the same (or not). Complexity is also a bit subjective, so again - let's have the discussion for now and then see how to tackle this (with an additional PR).

flask_test_app/app.py
merkata: I'd go with simplicity in that case - it's clear and does what it does. Using some ORM and then in tests might be too much, it would also hint at our preference (for instance sqlalchemy), where a developer should chose whatever they want, if they want to use an ORM in the first place.

src/charm.py
merkata: I think a pattern emerges from the need to solve something and then reusing that solution so may times that a pattern is formed, so there really should be some examples :) Then @yanksyoon  - could you elaborate here and link to some code? Thanks!

src/charm.py
merkata: Good point, I think it should block until data provided over the relation is valid.

flask_test_app/app.py
jdkandersson: I think we can leave the type hints for now, so this file can be excluded from mypy checks which should resolve the issue

flask_test_app/app.py
jdkandersson: the flask-sqlalchemy tool makes the sample app much simpler, you don't have to define models to use it. The queries for the version can be done I'm guessing using `db.engine.execute("<sql here>")` and these setup and teardown functions are not needed. This is the quick start for the library: https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/quickstart/#configure-the-extension

src/charm.py
nrobinaubertin: Outdated due to the fact that I moved most of the new code in its own module

src/charm.py
nrobinaubertin: Outdated due to the fact that I moved most of the new code in its own module

src/charm.py
nrobinaubertin: I decided to move most of the code into its own module. This should resolve this discussion

src/charm.py
nrobinaubertin: Done

flask_test_app/app.py
nrobinaubertin: After discussion with David, it was decided that I should go the `os.environ` route for now and that more discussions about it could arise.

weiiwang01: I am a little bit confused by the risk introduced by putting the database URI string into the flask configuration. I think if an attacker can access the flask configuration, the attacker should have access to the environment variable too since they are all unprotected in the memory.

flask_test_app/app.py
```diff
@@ -0,0 +1,112 @@
...
+            cursor.execute(sql)
+            cursor.fetchone()
+            return "SUCCESS"
+    return "FAIL"
+
+
+@app.route("/mysql/env")
+def mysql_env():
+    """Mysql env endpoint."""
+    return app.config.get("MYSQL_DB_CONNECT_STRING")
```
weiiwang01: Should this and the `postgresql_env` endpoint be replaced by the `config` endpoint?
For example, a request to `/mysql/env` can be replaced by `/config/MYSQL_DB_CONNECT_STRING`.
This way, we can save two endpoint functions.

src/charm.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, *args: typing.Any) -> None:
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
+        self.databases = Databases(charm=self)
```
weiiwang01: I think we can make this variable private.

src/charm.py
```diff
@@ -115,14 +117,16 @@ def flask_layer(self) -> dict:
         Returns:
             The pebble layer definition for flask application.
         """
+        environment = self._flask_app.flask_environment
+        environment.update(self.databases.get_uris())
```
weiiwang01: Should we catch and handle the `InvalidDatabaseRelationDataError` here?

src/databases.py
```diff
@@ -0,0 +1,108 @@
...
+        Returns:
+            DatabaseURI containing details about the data provider integration
+
+        Raises:
+            InvalidDatabaseRelationDataError: if the database relation has invalid data
+        """
+        db_uris: typing.Dict[str, str] = {}
+
+        # the database_requirer could not be defined
+        # if _database_uri() is called before its initialization
```
weiiwang01: Which is the `_database_uri` method?

flask_test_app/app.py
nrobinaubertin: To add to the discussion: even in flask documentation, there are examples of using `app.config` to store secrets: https://flask.palletsprojects.com/en/2.3.x/config/#configuring-from-environment-variables

flask_test_app/app.py
jdkandersson: We can leave it like this for now

src/charm.py
```diff
@@ -43,7 +44,7 @@ def __init__(self, *args: typing.Any) -> None:
             flask_container=self.unit.get_container(FLASK_CONTAINER_NAME),
         )
         self._flask_app = FlaskApp(charm_state=self._charm_state)
-        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.config_changed, self.on_config_changed)
```
jdkandersson: What is the reason for changing this? The event handlers should be private methods with a certain naming conventions, ISD014 specifies that in more detail

src/charm.py
```diff
@@ -136,75 +111,22 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
...
-        return env_config
-
     def flask_layer(self) -> dict:
         """Generate the pebble layer definition for flask application.
 
         Returns:
             The pebble layer definition for flask application.
         """
+        environment = self._flask_app.flask_environment
+        environment.update(self.databases.get_uris())
```
jdkandersson: Should we be consistent in whether the environment is returned as an attribute or as a function call?

src/charm.py
nrobinaubertin: I'm calling them from the database module (and I have to do that), so they are not private anymore

src/charm.py
nrobinaubertin: I find it strange that a property could raise an exception

flask_test_app/app.py
nrobinaubertin: I removed them (they were not necessary) but we cannot use the config endpoint since they are not available through `app.config`

src/charm.py
nrobinaubertin: Done

src/charm.py
nrobinaubertin: Done

src/databases.py
nrobinaubertin: Fixed

src/charm.py
jdkandersson: So should `.flask_environment` be changed to a function then?

src/databases.py
```diff
@@ -0,0 +1,108 @@
...
+        Returns:
+            DatabaseRequires object produced by the data_platform_libs.v0.data_interfaces library
+        """
+        database_requirer = DatabaseRequires(
+            self._charm,
+            relation_name=relation_name,
+            database_name=database_name,
+        )
+        self._charm.framework.observe(
+            database_requirer.on.database_created, self._charm.on_config_changed
```
jdkandersson: We discussed this for the Wordpress case as well, there shouldn't be these kind of circular dependencies between the Databases module and the charm. If the database relation change event needs to make changes at the charm level, then the observe initiation should be there, otherwise the event handling logic should only be on the Databases class.

Also, the ISD014 spec describes that there should be a 1-1 relationship between event and event handlers, so the `database_created` event should not use the `config_changed` event handler

If you would like to see an example of how a similar problem was solved, take a look at the Wordpress refactor spec or talk to Yanks 🙂

src/charm.py
jdkandersson: Ah I didn't notice that, I have left another comment to discuss that

src/databases.py
jdkandersson: This is also looking a lot like the observer pattern, it would be good not to re-invent it. We have discussed the Wordpress case during the sprint as a good way forward

nrobinaubertin: There's no simple way (IMHO) around the fact that we need to pebble replan to take changes to the flask layer into account.
Citing the current wp refactor spec: "Furthermore, if the event has changes that affect pebble configuration, the callback to the main charm class should be used". 

> This is also looking a lot like the observer pattern, it would be good not to re-invent it

I'm aware that it looks like it. But, as we discussed during our call, there's no accepted nor finalized version of the observer pattern for now. I'll be happy to change things in another PR but this seems out of scope for me right now.

jdkandersson: Given Gregory is back today, let's get him to take a look at this PR as well

jdkandersson: The other note, the Wordpress spec was discussed with Gregory and myself at the Prague sprint and these kind of issues came up and the spec was changed to address the issues. So just because it isn't finalised and approved doesn't mean that we should come up with other ways of doing the same thing which contains issues from an earlier version of the spec

src/charm_state.py
```diff
@@ -22,6 +20,9 @@
 from charm_types import WebserverConfig
 from exceptions import CharmConfigInvalidError
 
+if typing.TYPE_CHECKING:
+    from charm import FlaskCharm
```
gregory-schiano, line 24: Great, I think we should use `TYPE_CHECKING` more frequently

src/databases.py
gregory-schiano: I think we shouldn't keep a call back to the charm from the module as it creates circular dependency. My proposal is to do a step further without implementing the full `Observer pattern` as discussed with Niels and Yanks. Indeed the full pattern would require some refactoring on parts of the charm which aren't in the scope of this task. We'll have a further task which will ask to refactor using the observer pattern and a task to generalise with a parent class to get ready for the 12-factorization of this project.

My proposal is to remove the callback and manage the hook in this module by doing something like:
```python
def on_database_created(self, event: DatabaseCreatedEvent):
   container = self.charm.unit.get_container(FLASK_CONTAINER_NAME)
   plan = container.get_plan()
   plan.services["myservice"].environment.update(self.get_uris())
   container.replan()
 ```
 
 Which removes the callback to the charm, while not being the pattern as we want

File: src/constants.py [hunk 22]
```diff
@@ -6,3 +6,5 @@
 FLASK_CONTAINER_NAME = "flask-app"
 FLASK_SERVICE_NAME = "flask-app"
 FLASK_ENV_CONFIG_PREFIX = "FLASK_"
+FLASK_DATABASE_NAME = "flask-app"
+FLASK_SUPPORTED_DATABASES = ("mysql", "postgresql")
```
gregory-schiano: Wouldn't it be better to read that dynamically from the metadata ?
Don't think it'll require a lot of work while preparing for the next steps

src/databases.py
```diff
@@ -0,0 +1,108 @@
...
+            return db_uris
+
+        for database, db_requires in self._databases.items():
+            relation_data = list(db_requires.fetch_relation_data().values())
+
+            if not relation_data:
+                continue
+
+            # There can be only one database integrated at a time
+            # see: metadata.yaml
```
gregory-schiano: +`with the same relation name`. The comment could be confusing there could be multiple Mysql integrated, but they'll use different relation names

src/databases.py
```diff
@@ -0,0 +1,108 @@
...
+            data = relation_data[0]
+
+            # Check that the relation data is well formed according to the following json_schema:
+            # https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/mysql_client/v0/schemas/provider.json
+            if not all(data.get(key) for key in ("endpoints", "username", "password")):
```
gregory-schiano, line 138: Shouldn't we trust the `Provider` for this ? I'm a bit balanced @jdkandersson

src/databases.py
```diff
@@ -0,0 +1,108 @@
...
+                    f"Incorrect relation data from the data provider: {data}"
+                )
+
+            database_name = data.get("database", db_requires.database)
+            endpoint = data["endpoints"].split(",")[0]
+            db_uris[f"{database.upper()}_DB_CONNECT_STRING"] = (
+                f"{database}://"
+                f"{data['username']}:{data['password']}"
+                f"@{endpoint}/{database_name}"
+            )
```
gregory-schiano: Could be interesting if there's a library that modeled the DB connection string and use it instead of simply concatenating

gregory-schiano (commented):
I still need to review the tests

src/databases.py
nrobinaubertin: I could use [urllib.parse.urlunparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunparse)

src/databases.py
jdkandersson: In Prague we discussed the interfaces with Pietro and a few others from the observability team and it actually seems common practice to validate the data received over the relation interface. Either we can have unhandled exceptions in cases where the data is unexpected or we can handle that a bit more gracefully. It would actually be good for the charm libraries to give you objects with the data and the library of the interface to provide the checks and de-serialisation

Eventually the interface repo should have Pydantic models in them that we can just pull in so we won't have to maintain our own. Could be worth checking if they are available for the database already

src/constants.py
nrobinaubertin: Done

src/databases.py
nrobinaubertin: Actually it's not really simple to use `urlunparse` and that will still need some string concatenating.  
There's the [`URL` class of sqlalchemy](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/engine/url.py#L45) but I don't want to import it just for that.

src/databases.py
```diff
@@ -6,37 +6,67 @@
...
         """Initialize a new instance of the Databases class.
 
         Args:
             charm: The main charm. Used for events callbacks
         """
+        # The following is necessary to be able to subscribe to callbacks from ops.framework
+        super().__init__(charm, "databases")
         self._charm = charm
+
+        with open("metadata.yaml", encoding="utf-8") as metadata_fo:
```
jdkandersson: pathlib provides an API `.read_text` which makes this much easier: https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

src/databases.py
gregory-schiano: Let's keep it as it is for now, we'll see later if such lib is needed

src/databases.py
gregory-schiano: They seem to have a library helper to create pydantic data models, but didn't find any data model effectively implemented
Also I didn't realise that we weren't taking data from an event (which is emitted by the requirer library and should contain validated data) but it comes from the relation databag directly which doesn't have any validation.
So all good

tests/unit/test_databases.py
```diff
@@ -0,0 +1,116 @@
...
+    for relation in relations:
+        database_charm_name = f"some_db_charm_{relation['interface']}"
+        relation_id: int = harness.add_relation(relation["interface"], database_charm_name)
+        harness.add_relation_unit(relation_id, f"{database_charm_name}/0")
+        harness.update_relation_data(relation_id, database_charm_name, relation["data"])
+        harness.update_config()
+
+    # Allowing protected access to test the output
+    # pylint: disable=protected-access
+    assert harness.charm._databases.get_uris() == expected_output
```
gregory-schiano: Isn't it easier and more `unit` to mock fetch_relation_data ?
This looks more like a functional test

gregory-schiano (commented):
Reviewed the tests

src/charm.py
nrobinaubertin: Done

src/databases.py
nrobinaubertin: Fixed

src/databases.py
nrobinaubertin: Changed

tests/unit/test_databases.py
nrobinaubertin: Changed to use a mock

src/databases.py
nrobinaubertin: Reworked

src/databases.py
```diff
@@ -0,0 +1,138 @@
...
+        Args:
+            charm: The main charm. Used for events callbacks
+        """
+        # The following is necessary to be able to subscribe to callbacks from ops.framework
+        super().__init__(charm, "databases")
+        self._charm = charm
+
+        metadata = yaml.safe_load(pathlib.Path("metadata.yaml").read_text(encoding="utf-8"))
+        self._db_interfaces = (
+            name for name in list(metadata["requires"]) if name.startswith("db_")
```
gregory-schiano: Instead of looking for a prefix I'd rather look for requires that uses `postgres_client` or `mysql_client` as interface, which would make it more generic

nrobinaubertin: I understand. Is it not possible that we use those interfaces for something else than the "db" ?
It's probably too much of an edge case now that I think of it but I thought that we could use postgres/mysql for something cache related or something else. In that case we could have those interfaces used in another way than intended.

nrobinaubertin:
I'm waiting for https://github.com/canonical/flask-k8s-operator/pull/20 before merging this PR (if everything is solved).

src/databases.py
gregory-schiano: Whatever the reason is, everytime we have a postgres_client or mysql_client interface used, we need to instanciate the requirer to manage the relation.
That's why I'm suggesting this change

nrobinaubertin: Fixed

src/charm.py
nrobinaubertin: Changed to a function

weiiwang01 (dismissed):

gregory-schiano:
I think the branch should be rebased as the diff shows diff from other PRs

nrobinaubertin:
> I think the branch should be rebased as the diff shows diff from other PRs

It would have been difficult to do that since multiple commits are merged commits. I reviewed the PR to fix issues and remove left overs from merges.

gregory-schiano (approved):
LGTM

merkata (approved):

PR #18, opened 2023-05-30
Add sentinel image and sentinel image check

weiiwang01:
The flask-k8s charm operates in the "Bring Your Own Container Image" mode, where it's expected that users provide a custom container image for their Flask application during deployment.

However, to conform to CharmHub's requirements, all resources defined in the charm metadata must be supplied during the release of the charm to CharmHub. To resolve this, a sentinel image is used as a placeholder in the charm metadata. This sentinel image does not host a Flask application but is merely used to fulfill the requirement set by CharmHub.

To accommodate the change, checks are added to the charm to detect the absence of a user-provided container image, allowing it to enter block state and provide an informative error message, guiding the user to redeploy the charm with their Flask application container image.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,6 +18,7 @@ from constants import FLASK_CONTAINER_NAME, FLASK_SERVICE_NAME
 from exceptions import CharmConfigInvalidError, PebbleNotReadyError
 from flask_app import FlaskApp
 from observability import Observability
+from sentinel import Sentinel
 from webserver import GunicornWebserver
 
 logger = logging.getLogger(__name__)
@@ -33,6 +34,9 @@ class FlaskCharm(CharmBase):
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
+        self._sentinel = Sentinel(charm=self)
+        if not self._precondition_check():
+            return
         try:
             self._charm_state = CharmState.from_charm(charm=self)
         except CharmConfigInvalidError as exc:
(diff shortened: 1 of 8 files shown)


mthaddon:
Could we use a rockcraft.yaml for this image rather than a Dockerfile?

weiiwang01:
> Could we use a rockcraft.yaml for this image rather than a Dockerfile?

This is a very special image that only contain an empty file `/null` and nothing else, I am a little reluctant to use rockcraft since rockcraft will add pebble to the final image.

src/charm.py
```diff
@@ -59,6 +63,21 @@ def __init__(self, *args: typing.Any) -> None:
...
+    def _precondition_check(self) -> bool:
+        """Check preconditions for charm execution are met.
+
+        Returns: True if the precondition is met, False otherwise.
+        """
+        if self._sentinel.is_sentinel_image():
+            self._update_app_and_unit_status(
+                BlockedStatus(
+                    "charm requires a Flask app image, "
+                    "redeploy with '--resource flask-app-image=<your-image>'"
```
mthaddon, line 75: I think you can also use `attach-resource` to do this without having to redeploy. I realise we have limited space here in terms of the message we display, but giving people a way to get out of this without needing to redeploy seems like it might be worthwhile.

mthaddon:
> > Could we use a rockcraft.yaml for this image rather than a Dockerfile?
> 
> This is a very special image that only contain an empty file `/null` and nothing else, I am a little reluctant to use rockcraft since rockcraft will add pebble to the final image.

This may need discussion/agreement with some others as well. I think it's probably better for us to standardise on ROCKs than Dockerfiles, even for something as trivial as this.

src/charm.py
weiiwang01: Thanks for the info! @nrobinaubertin has suggested we use a default image that hosts a welcome page containing tips for the next step. I added the `attach-resource` command to that welcome page.

Here's the new pull request #19

weiiwang01:
> > > Could we use a rockcraft.yaml for this image rather than a Dockerfile?
> > 
> > 
> > This is a very special image that only contain an empty file `/null` and nothing else, I am a little reluctant to use rockcraft since rockcraft will add pebble to the final image.
> 
> This may need discussion/agreement with some others as well. I think it's probably better for us to standardise on ROCKs than Dockerfiles, even for something as trivial as this.

The sentinel image solution has been replaced by the default image solution. Close this pull request in favor of #19.

PR #20, opened 2023-05-30
Rework fixtures to avoid unnecessary deployments

nrobinaubertin:
This is a rework of the fixtures in order to avoid unnecessary deployments for some of the integration tests.  
The goal is to try to reduce the number of CI timeouts and if possible to speed things up.

The speedup seems to be real (from around 30mn to 15mn) but this will need confirmation in the long run.

### DIFF
--- a/tests/integration/test_charm.py
+++ b/tests/integration/test_charm.py
@@ -166,6 +166,7 @@ async def test_with_ingress(
     ops_test: OpsTest,
     model: Model,
     flask_app: Application,
+    traefik_app,  # pylint: disable=unused-argument
     traefik_app_name: str,
     external_hostname: str,
     get_unit_ips: typing.Callable[[str], typing.Awaitable[tuple[str, ...]]],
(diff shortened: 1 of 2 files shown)


gregory-schiano (approved):
LGTM

arturo-seijas (approved):
LGTM

tests/integration/test_charm.py
```diff
@@ -166,6 +166,7 @@ async def test_with_ingress(
     ops_test: OpsTest,
     model: Model,
     flask_app: Application,
+    traefik_app,  # pylint: disable=unused-argument
```
jdkandersson, line 169: Instead of disabling unused arguments, on tests (not on fixtures), `pytest.mark.usefixtures(...)` can be used: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures

PR #21, opened 2023-06-02
Add webserver WSGI path configuration

weiiwang01:
And a charm configuration called `webserver_wsgi_path` to allow the charm run with different Flask project structures.


config.yaml
```diff
@@ -58,3 +58,7 @@ options:
   webserver_workers:
     type: int
     description: The number of webserver worker processes for handling requests.
+  webserver_wsgi_path:
+    type: string
+    default: "app:app"
+    description: The WSGI application path. By default, it's set to "app:app".
```
arturo-seijas: Missing line at EOF

weiiwang01: Added, thanks!

arturo-seijas (dismissed):
LGTM

config.yaml
```diff
@@ -58,3 +58,7 @@ options:
   webserver_workers:
     type: int
     description: The number of webserver worker processes for handling requests.
+  webserver_wsgi_path:
```
gregory-schiano, line 61: I'm wondering if we shouldn't name this `webserver_entrypoint` as it's more generic/reusable and path could look like FS path

weiiwang01: WSGI path is a common technical term in the Flask domain, so I think it might be more understandable for Flask users.

gregory-schiano: Yeap but we're in the webserver module, which is meant to become generic

weiiwang01: I think the WSGI path is applicable for all Flask web servers, such as Gunicorn, uWSGI, etc.

gregory-schiano: The webserver doesn't aim to be Python specific, it aims to be generic to any 12-factor app charm as any web application requires a webserver
Using generic config allows to have a better user experience. We could keep this name for now, but we'll very likely rename it later, so it doesn't cost much to name is appropriately now

weiiwang01: The problem is that wsgi path is specify to the WSGI server, for other type of web servers such as fastcgi server, there's a correlating concept to the wsgi path in WSGI server. Maybe we can rename the `webserver` module to `wsgi_webserver` module?

weiiwang01: For other configurations, there are similar problems, for example, a cgi webserver doesn't have the concept of workers or threads.

src/charm_state.py
```diff
@@ -115,11 +117,15 @@ def __init__(
                 or None if not specified.
             webserver_timeout: The request silence timeout for the web server,
                 or None if not specified.
+            webserver_wsgi_path: The WSGI application path, or None if not specified.
```
amandahla, line 121: None or "app:app"?

weiiwang01: The default value for the `webserver_wsgi_path` function argument is `None`. This is due to the typing of the ops framework. The `None` input will be interpreted as `"app:app"`.

config.yaml
jdkandersson: The renaming later sounds like a good option when we have another example, e.g., the fastAPI or spring boot case

jdkandersson (dismissed):

merkata (dismissed):

amandahla (approved):
LGTM

jdkandersson (approved):

PR #23, opened 2023-06-12
Use charm managed secret key as default flask SECRET_KEY configuration

weiiwang01:
This pull request enables the flask-k8s charm to generate its a secret key upon initial deployment, which is used as the default Flask `SECRET_KEY` configuration if the user does not provide a `flask_secret_key` configuration. Additionally, a charm action called `rotate-secret-key` is introduced to allow resetting the charm-managed secret value, providing a safeguard in the event of a data breach.


actions.yaml
```diff
@@ -0,0 +1,6 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+rotate-secret-key:
+  description: >
+    rotating the flask secret key. Users will be forced to log in again.
```
arturo-seijas: s/rotating/Rotates?

actions.yaml
```diff
@@ -0,0 +1,6 @@
...
+    This might be useful under security breach circumstances.
```
arturo-seijas: s/security breach circumstances/is a security breach occurs? I'm thinking about keeping it as simple as possible for documentation purposes

src/charm.py
```diff
@@ -100,19 +107,34 @@ def container(self) -> Container:
         container = self.unit.get_container(FLASK_CONTAINER_NAME)
         return container
 
-    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
-        """Configure the flask pebble service layer.
+    @property
+    def _is_precondition_satisfied(self) -> bool:
+        """Check if the precondition for the Flask application has been satisfied.
```
arturo-seijas, line 122: Preconditions for what? Can you specify?

src/charm.py
```diff
@@ -173,6 +207,22 @@ def _on_statsd_prometheus_exporter_pebble_ready(self, _event: PebbleReadyEvent)
         statsd_container.add_layer("statsd-prometheus-exporter", statsd_layer, combine=True)
         statsd_container.replan()
 
+    def _on_rotate_secret_key_action(self, event: ActionEvent) -> None:
+        """Handle the rotate-secret-key action.
+
+        Args:
+            event: the action event that trigger this callback.
```
arturo-seijas, line 224: s/trigger/triggered

src/charm.py
gregory-schiano: Yes I'd rename it to `_flask_app_precondition_satisfied`

src/charm.py
```diff
@@ -100,19 +107,34 @@ def container(self) -> Container:
...
+            return False
+        if not self._secret_storage.is_initialized:
+            logger.info("secret storage is not initialized, defer config-changed")
+            return False
+        return True
+
+    def _restart_flask_application(self) -> None:
+        """Start or restart the flask application with the latest charm configuration.
+
+        The flask charm must be ready (i.e. preconditions meet) before calling this function.
```
gregory-schiano, line 144: If the preconditions must be met, then we should check them in this method and raise an exception if they're not met

src/flask_app.py
```diff
@@ -6,18 +6,22 @@
...
+    def __init__(self, charm_state: CharmState, secret_storage: SecretStorage):
         """Initialize a new instance of the FlaskApp class.
 
         Args:
             charm_state: The state of the charm that the FlaskApp instance belongs to.
+            secret_storage: The secret storage object associated with the Flask charm, the secret
+                storage must be initialized before calling any methods of the FlaskApp class.
         """
         self._charm_state = charm_state
+        self._secret_storage = secret_storage
```
gregory-schiano: It indirectly makes the app depandant of the charm which I'm not fan
Why not having the secret_key as part of the state ?

gregory-schiano (commented):
Didn't look deeper in the PR
I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
This will avoid a dependency between the Flask Service and the SecretStorageService

src/charm.py
arturo-seijas: I'm not sure about that. From what I can see, not meeting those preconditions would be part of the normal flow, for instance, when pebble is not ready

src/charm.py
arturo-seijas: I also miss some additional detail in the docstring. I assume we are talking about preconditions for startup, but it's a bit unclear unless you go through the code

actions.yaml
weiiwang01: Updated, thanks!

actions.yaml
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: The preconditions are listed in the following paragraph.

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
weiiwang01: If the preconditions aren't met, the charm should defer the event. But the `_restart_flask_application` doesn't have any context of the event, so the check need to be done outside.

weiiwang01:
> Didn't look deeper in the PR
> I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
> This will avoid a dependency between the Flask Service and the SecretStorageService

The problem of putting the secret key into the charm state is that the secret key can not be directly derived from the charm. This means to create the charm state from the charm, we need to pass the secret storage instance along with the charm to the `from_charm` factory function.

weiiwang01:
> > Didn't look deeper in the PR
> > I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
> > This will avoid a dependency between the Flask Service and the SecretStorageService
> 
> The problem of putting the secret key into the charm state is that the secret key can not be directly derived from the charm. This means to create the charm state from the charm, we need to pass the secret storage instance along with the charm to the `from_charm` factory function.

The `SecretStorage` is more like `DatabaseRequires` or `GrafanaDashboardProvider`, and we are planning to extract the `SecretStorage` to a charm lib. So it's not very closely related to the flask-k8s charm.

gregory-schiano:
> > > Didn't look deeper in the PR
> > > I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
> > > This will avoid a dependency between the Flask Service and the SecretStorageService
> > 
> > 
> > The problem of putting the secret key into the charm state is that the secret key can not be directly derived from the charm. This means to create the charm state from the charm, we need to pass the secret storage instance along with the charm to the `from_charm` factory function.
> 
> The `SecretStorage` is more like `DatabaseRequires` or `GrafanaDashboardProvider`, and we are planning to extract the `SecretStorage` to a charm lib. So it's not very closely related to the flask-k8s charm.

I get your point, but the `Requirer` or `Provider` classes aren't supposed to be used by services, they're autonomous, therefore that causes no problem.
Here that's different, a service is making use of it, leading to an indirect dependency between the charm and that service.

My initial proposal wasn't to pass the SecretStorage library to the state through `from_charm` but to pass the state to the SecretStorage library in the constructor and let the SecretStorage service change the secretKey attribute in the state depending of what's hapenning.
The state doesn't represent the charm state but the Workload state (hence my willing to name it `WorkloadState`), and the secretKey is part of it.

I think that's an architectural element we should discuss, which is not described in the `Managing charm complexity` document. How are all these modules supposed to work with eachother
In my opinion, we must distinguish 3 type of modules:
- Workload State: Must not depend on the charm, is initialized in charm constructor, implements a `from_charm` method to initialize, don't depend on a service or a library
- Charm Service: Can depend on the container, is instanciated in charm constructure, can mutate the Workload State (therefore depend on it), don't depend on a library
- Charm libraries: Must depend on the charm (to attach hooks to events), is instanciated in charm constructure, don't depend on a service or a library

What do you think ? @jdkandersson @weiiwang01

weiiwang01:
> > > > Didn't look deeper in the PR
> > > > I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
> > > > This will avoid a dependency between the Flask Service and the SecretStorageService
> > > 
> > > 
> > > The problem of putting the secret key into the charm state is that the secret key can not be directly derived from the charm. This means to create the charm state from the charm, we need to pass the secret storage instance along with the charm to the `from_charm` factory function.
> > 
> > 
> > The `SecretStorage` is more like `DatabaseRequires` or `GrafanaDashboardProvider`, and we are planning to extract the `SecretStorage` to a charm lib. So it's not very closely related to the flask-k8s charm.
> 
> I get your point, but the `Requirer` or `Provider` classes aren't supposed to be used by services, they're autonomous, therefore that causes no problem. Here that's different, a service is making use of it, leading to an indirect dependency between the charm and that service.
> 
> My initial proposal wasn't to pass the SecretStorage library to the state through `from_charm` but to pass the state to the SecretStorage library in the constructor and let the SecretStorage service change the secretKey attribute in the state depending of what's hapenning. The state doesn't represent the charm state but the Workload state (hence my willing to name it `WorkloadState`), and the secretKey is part of it.
> 
> I think that's an architectural element we should discuss, which is not described in the `Managing charm complexity` document. How are all these modules supposed to work with eachother In my opinion, we must distinguish 3 type of modules:
> 
> * Workload State: Must not depend on the charm, is initialized in charm constructor, implements a `from_charm` method to initialize, don't depend on a service or a library
> * Charm Service: Can depend on the container, is instanciated in charm constructure, can mutate the Workload State (therefore depend on it), don't depend on a library
> * Charm libraries: Must depend on the charm (to attach hooks to events), is instanciated in charm constructure, don't depend on a service or a library
> 
> What do you think ? @jdkandersson @weiiwang01

If we passing `CharmState` to `SecretStorage` and modifying the `CharmState`, this could have two potential problems. First this makes the `CharmState` mutable, which we want to avoid. Also, the end goal of `SecretStorage` is to extract it to a charm library, thus be useful for other charms such as the wordpress-k8s charm. Making `SecretStorage` modifying the `CharmState` creates a dependency between the flask's `CharmState` and `SecretStorage` which will reduce the generosity of the `SecretStorage`.

src/charm.py
```diff
@@ -66,6 +65,14 @@ def __init__(self, *args: typing.Any) -> None:
         )
         self._observability = Observability(charm=self, charm_state=self._charm_state)
 
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(
+            self.on.statsd_prometheus_exporter_pebble_ready,
+            self._on_statsd_prometheus_exporter_pebble_ready,
+        )
+        self.framework.observe(self.on.rotate_secret_key_action, self._on_rotate_secret_key_action)
+        self.framework.observe(self.on.secret_storage_relation_changed, self._on_config_changed)
```
jdkandersson, line 84: Each event should have its own handler. They can delegate to shared methods for code re-usability

src/secret_storage.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Attrs:
+        is_initialized: True if the SecretStorage has been initialized.
+    """
+
+    # bandit think this is a password
+    _FLASK_SECRET_KEY_KEY = "flask_secret_key"  # nosec
+    _PEER_RELATION_NAME = "secret-storage"
+
+    def __init__(self, charm: "FlaskCharm"):
```
jdkandersson: This is a circular dependency, should the type hint be `CharmBase`?

src/secret_storage.py
```diff
@@ -0,0 +1,94 @@
...
+
+class SecretStorage(Object):
+    """A class that manages secret keys required by the FlaskCharm.
+
+    Attrs:
+        is_initialized: True if the SecretStorage has been initialized.
+    """
+
+    # bandit think this is a password
+    _FLASK_SECRET_KEY_KEY = "flask_secret_key"  # nosec
```
jdkandersson, line 20: Could these be module level constants?

jdkandersson:
> > > > > Didn't look deeper in the PR
> > > > > I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
> > > > > This will avoid a dependency between the Flask Service and the SecretStorageService
> > > > 
> > > > 
> > > > The problem of putting the secret key into the charm state is that the secret key can not be directly derived from the charm. This means to create the charm state from the charm, we need to pass the secret storage instance along with the charm to the `from_charm` factory function.
> > > 
> > > 
> > > The `SecretStorage` is more like `DatabaseRequires` or `GrafanaDashboardProvider`, and we are planning to extract the `SecretStorage` to a charm lib. So it's not very closely related to the flask-k8s charm.
> > 
> > 
> > I get your point, but the `Requirer` or `Provider` classes aren't supposed to be used by services, they're autonomous, therefore that causes no problem. Here that's different, a service is making use of it, leading to an indirect dependency between the charm and that service.
> > My initial proposal wasn't to pass the SecretStorage library to the state through `from_charm` but to pass the state to the SecretStorage library in the constructor and let the SecretStorage service change the secretKey attribute in the state depending of what's hapenning. The state doesn't represent the charm state but the Workload state (hence my willing to name it `WorkloadState`), and the secretKey is part of it.
> > I think that's an architectural element we should discuss, which is not described in the `Managing charm complexity` document. How are all these modules supposed to work with eachother In my opinion, we must distinguish 3 type of modules:
> > 
> > * Workload State: Must not depend on the charm, is initialized in charm constructor, implements a `from_charm` method to initialize, don't depend on a service or a library
> > * Charm Service: Can depend on the container, is instanciated in charm constructure, can mutate the Workload State (therefore depend on it), don't depend on a library
> > * Charm libraries: Must depend on the charm (to attach hooks to events), is instanciated in charm constructure, don't depend on a service or a library
> > 
> > What do you think ? @jdkandersson @weiiwang01
> 
> If we passing `CharmState` to `SecretStorage` and modifying the `CharmState`, this could have two potential problems. First this makes the `CharmState` mutable, which we want to avoid. Also, the end goal of `SecretStorage` is to extract it to a charm library, thus be useful for other charms such as the wordpress-k8s charm. Making `SecretStorage` modifying the `CharmState` creates a dependency between the flask's `CharmState` and `SecretStorage` which will reduce the generosity of the `SecretStorage`.

The `SecretStorage` service is lower level than a database service since it is a way to remember state across invocations of the charm. It is quite similar to the charm configuration in that way, which is why it is ok to pass it or its values to the constructor of `CharmState`. It wouldn't be ok for a database service since that in turn could depend on `CharmState` thus creating a circular dependency. This generalises to `<any>Storage` as being a way to remember values across invocations. There are arguments for passing `SecretStorage` or just the secret value to `CharmState`. Passing the secret value in `from_charm` decouples the state from `SecretStorage`, though calls to rotate the secret would need to update the state after the charm has been initialised. Passing `SecretStorage` in `from_charm` would mean that the value of the secret key is kept up to date even after the secret is rotated. The latter sounds like it would reduce the bug rate since the developer doesn't have to remember to re-create the state after rotating the secret. It is also aligned with that the state can depend on configuration and `SecretStorage` is essentially charm generated configuration.

Slightly modified categorisations from what Gregory wrote:
* State storage: Used to keep track of values across charm invocations if it isn't possible in other ways, such as via configuration or juju secrets
* Workload State: Must not depend on the charm, is initialised in charm constructor using values from the configuration, secrets and storage services, implements a `from_charm` method to initialise, don't depend on a service or a library
* Charm Service: Can depend on the container (for the purpose of file system interactions and command execution), is instantiated in charm constructer, can depend on the state, don't depend on a library
* Charm libraries: Must depend on the charm (to attach hooks to events), is instanciated in charm constructure, don't depend on a service or a library

I'm not sure I follow this part yet, perhaps better for a separate discussion: can mutate the Workload State (therefore depend on it) - the state is currently not mutable directly, i.e., we shouldn't update values directly using `state.<attr> = <value>`. The state may have properties which retrieve values at access time. If the retrieval results in a different value, then the attribute will have a different value

src/charm.py
weiiwang01: Updated, thanks!

src/secret_storage.py
weiiwang01: Updated, thanks!

src/secret_storage.py
weiiwang01: Since the module only have one class, I think putting in the class is okay?

weiiwang01:
> > > > > > Didn't look deeper in the PR
> > > > > > I think the generated key should be in the state and the state passed to the SecretStorageService which will update the secretKey in the state directly
> > > > > > This will avoid a dependency between the Flask Service and the SecretStorageService
> > > > > 
> > > > > 
> > > > > The problem of putting the secret key into the charm state is that the secret key can not be directly derived from the charm. This means to create the charm state from the charm, we need to pass the secret storage instance along with the charm to the `from_charm` factory function.
> > > > 
> > > > 
> > > > The `SecretStorage` is more like `DatabaseRequires` or `GrafanaDashboardProvider`, and we are planning to extract the `SecretStorage` to a charm lib. So it's not very closely related to the flask-k8s charm.
> > > 
> > > 
> > > I get your point, but the `Requirer` or `Provider` classes aren't supposed to be used by services, they're autonomous, therefore that causes no problem. Here that's different, a service is making use of it, leading to an indirect dependency between the charm and that service.
> > > My initial proposal wasn't to pass the SecretStorage library to the state through `from_charm` but to pass the state to the SecretStorage library in the constructor and let the SecretStorage service change the secretKey attribute in the state depending of what's hapenning. The state doesn't represent the charm state but the Workload state (hence my willing to name it `WorkloadState`), and the secretKey is part of it.
> > > I think that's an architectural element we should discuss, which is not described in the `Managing charm complexity` document. How are all these modules supposed to work with eachother In my opinion, we must distinguish 3 type of modules:
> > > 
> > > * Workload State: Must not depend on the charm, is initialized in charm constructor, implements a `from_charm` method to initialize, don't depend on a service or a library
> > > * Charm Service: Can depend on the container, is instanciated in charm constructure, can mutate the Workload State (therefore depend on it), don't depend on a library
> > > * Charm libraries: Must depend on the charm (to attach hooks to events), is instanciated in charm constructure, don't depend on a service or a library
> > > 
> > > What do you think ? @jdkandersson @weiiwang01
> > 
> > 
> > If we passing `CharmState` to `SecretStorage` and modifying the `CharmState`, this could have two potential problems. First this makes the `CharmState` mutable, which we want to avoid. Also, the end goal of `SecretStorage` is to extract it to a charm library, thus be useful for other charms such as the wordpress-k8s charm. Making `SecretStorage` modifying the `CharmState` creates a dependency between the flask's `CharmState` and `SecretStorage` which will reduce the generosity of the `SecretStorage`.
> 
> The `SecretStorage` service is lower level than a database service since it is a way to remember state across invocations of the charm. It is quite similar to the charm configuration in that way, which is why it is ok to pass it or its values to the constructor of `CharmState`. It wouldn't be ok for a database service since that in turn could depend on `CharmState` thus creating a circular dependency. This generalises to `<any>Storage` as being a way to remember values across invocations. There are arguments for passing `SecretStorage` or just the secret value to `CharmState`. Passing the secret value in `from_charm` decouples the state from `SecretStorage`, though calls to rotate the secret would need to update the state after the charm has been initialised. Passing `SecretStorage` in `from_charm` would mean that the value of the secret key is kept up to date even after the secret is rotated. The latter sounds like it would reduce the bug rate since the developer doesn't have to remember to re-create the state after rotating the secret. It is also aligned with that the state can depend on configuration and `SecretStorage` is essentially charm generated configuration.
> 
> Slightly modified categorisations from what Gregory wrote:
> 
> * State storage: Used to keep track of values across charm invocations if it isn't possible in other ways, such as via configuration or juju secrets
> * Workload State: Must not depend on the charm, is initialised in charm constructor using values from the configuration, secrets and storage services, implements a `from_charm` method to initialise, don't depend on a service or a library
> * Charm Service: Can depend on the container (for the purpose of file system interactions and command execution), is instantiated in charm constructer, can depend on the state, don't depend on a library
> * Charm libraries: Must depend on the charm (to attach hooks to events), is instanciated in charm constructure, don't depend on a service or a library
> 
> I'm not sure I follow this part yet, perhaps better for a separate discussion: can mutate the Workload State (therefore depend on it) - the state is currently not mutable directly, i.e., we shouldn't update values directly using `state.<attr> = <value>`. The state may have properties which retrieve values at access time. If the retrieval results in a different value, then the attribute will have a different value

Refactored the charm to embed the `SecretStorage` into `CharmState`, thanks!

src/flask_app.py
```diff
@@ -50,5 +46,5 @@ def flask_environment(self) -> dict[str, str]:
         }
         secret_key_env = f"{FLASK_ENV_CONFIG_PREFIX}SECRET_KEY"
         if secret_key_env not in env:
-            env[secret_key_env] = self._secret_storage.get_flask_secret_key()
+            env[secret_key_env] = self._charm_state.secret_storage.get_flask_secret_key()
```
jdkandersson: Should this be abstracted, e.g., `CharmState` has a `secret_key` property which calls `secret_storage.get_flask_secret_key()` since otherwise the `CharmState` needs to have a mock implementation of `SecretStorage` for testing

weiiwang01: The ops testing harness will automatically provide simulation for peer relation, so we don't need to mock the `SecretStorage` in unit tests ourself, just `harness.begin_with_initial_hooks()` is enough.

jdkandersson: The thing is that the aim is to be able to test modules without needing the harness, calling the function means that any modules that depend on the secret store will require to have unit tests with the harness

gregory-schiano: I agree, and have a similar comment for L134 and L230 in charm.py
I think that the charmstate shouldn't store the whole SecretStorage, but just the `get_flask_secret_key()` method
Calling a secretstorage method through the charmstate seems strange to me

What do you think @jdkandersson .

weiiwang01: The `SecretStorage` doesn't just store the secret key, there's also a method called `is_initialized` to check if the `SecretStorage` is initialized. Read or write the `SecretStorage` is an error before initialization.

jdkandersson: I think it is ok to have the `SecretStore` in the state to the degree necessary which could be just a record of the method or the entire thing, as long as the interactions are abstracted by the state

jdkandersson (approved):
Thanks!

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM, let's merge this :)

PR #26, opened 2023-06-28
Miscellaneous enhancements based on review comments

weiiwang01:
Following an internal review of the charm, several suggestions for improvement were identified. This pull request addresses those suggestions by implementing a minor refactor of the charm. No changes were made to the charm's features.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -71,7 +71,9 @@ class FlaskCharm(ops.CharmBase):
             self._on_statsd_prometheus_exporter_pebble_ready,
         )
         self.framework.observe(self.on.rotate_secret_key_action, self._on_rotate_secret_key_action)
-        self.framework.observe(self.on.secret_storage_relation_changed, self._on_config_changed)
+        self.framework.observe(
+            self.on.secret_storage_relation_changed, self._on_secret_storage_relation_changed
+        )
 
     def _update_app_and_unit_status(self, status: ops.StatusBase) -> None:
         """Update the application and unit status.
@@ -215,7 +217,7 @@ class FlaskCharm(ops.CharmBase):
         event.set_results({"status": "success"})
         self._restart_flask_application()
 
-    def _secret_storage_relation_changed(self, event: ops.RelationEvent) -> None:
+    def _on_secret_storage_relation_changed(self, event: ops.RelationEvent) -> None:
         """Handle the secret-storage-relation-changed event.
 
         Args:
--- a/src/charm.py
+++ b/src/charm.py
@@ -8,18 +8,8 @@ import logging
 import shlex
 import typing
 
+import ops
 from charms.traefik_k8s.v1.ingress import IngressPerAppRequirer
-from ops import (
-    ActionEvent,
-    ActiveStatus,
-    BlockedStatus,
-    CharmBase,
-    Container,
-    EventBase,
-    PebbleReadyEvent,
-    RelationEvent,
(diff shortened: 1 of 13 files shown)


gregory-schiano:
Can you write a description ?

src/databases.py
```diff
@@ -7,10 +7,9 @@
 import pathlib
 import typing
 
-import ops.framework
 import yaml
 from charms.data_platform_libs.v0.data_interfaces import DatabaseRequires, DatabaseRequiresEvent
-from ops.charm import CharmBase
+from ops import CharmBase, Object
```
gregory-schiano: I think the recommendation we to simply import ops and use the first-level object available

weiiwang01:
> Can you write a description ?

👍

tests/unit/test_charm_state.py
```diff
@@ -77,13 +76,13 @@ def test_charm_state_invalid_flask_config(harness: Harness, charm_config: dict)
...
-    "charm_config, wsgi_path",
-    [
-        pytest.param({}, "app:app", id="default"),
-        pytest.param({"webserver_wsgi_path": "foo.bar:app2"}, "foo.bar:app2", id="non-default"),
-    ],
-)
+CHARM_STATE_WSGI_PATH_TEST_PARAMS = [
+    pytest.param({}, "app:app", id="default"),
+    pytest.param({"webserver_wsgi_path": "foo.bar:app2"}, "foo.bar:app2", id="non-default"),
+]
```
gregory-schiano: I don't think that was required here as there aren't so many params. The comment was more for long parametrize calls

gtrkiller (dismissed):
LGTM

src/databases.py
```diff
@@ -49,7 +48,7 @@ def __init__(self, charm: CharmBase):
             for require in metadata["requires"].values()
             if require["interface"] in FLASK_SUPPORTED_DB_INTERFACES
         )
-
+        # automatically create database relation requirers to manage database relations
```
jdkandersson, line 51: Perhaps expand this a bit more to indicate why a dictionary is used?

src/databases.py
jdkandersson: I think either is fine, it depends on how much we actually need from it and whether the namespace is useful

tests/unit/test_charm_state.py
weiiwang01: Updated, thanks!

src/databases.py
weiiwang01: I changed to use the `ops` module directly.

jdkandersson (dismissed):

src/databases.py
weiiwang01: Added, thanks!

gregory-schiano (dismissed):
LGTM

src/charm.py
```diff
@@ -83,7 +73,7 @@ def __init__(self, *args: typing.Any) -> None:
         self.framework.observe(self.on.rotate_secret_key_action, self._on_rotate_secret_key_action)
         self.framework.observe(self.on.secret_storage_relation_changed, self._on_config_changed)
```
gregory-schiano: This needs to be changed, it should map to a dedicated method

jdkandersson (approved):

src/charm.py

gregory-schiano (approved):
LGTM

PR #28, opened 2023-07-06
Add support for the nginx-route relation

weiiwang01:
Add support for the `nginx-route` relation to the `flask-k8s` charm. Also, enable the automatic publish workflow to publish the charm to charmhub regularly.


src/charm.py
```diff
@@ -74,6 +75,16 @@ def __init__(self, *args: typing.Any) -> None:
...
             self.on.secret_storage_relation_changed, self._on_secret_storage_relation_changed
         )
+        self._require_nginx_route()
+
+    def _require_nginx_route(self) -> None:
+        """Set up the requirer side of the nginx-route relation."""
+        require_nginx_route(
+            charm=self,
+            service_name=self.unit.name,
+            service_hostname=self.unit.name,
```
mthaddon: I think you'd want `self.app.name` for both of these rather than `self.unit.name` as that would then be different for each unit.

weiiwang01: That's right, fixed, thanks!

mthaddon (dismissed):
LGTM

gregory-schiano:
I don't think we should support nginx-route on this charm and on the 12-factor globally
I'd rather implement as discussed the `ingress-to-nginx-route` or add `ingress` support on nginx-ingress-integrator

.github/workflows/publish_charm.yaml
```diff
@@ -0,0 +1,13 @@
+name: Publish to edge
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  test-and-publish-charm:
```
arturo-seijas: Picky but there' no testing here anymore. How about `publish-charm`?

metadata.yaml
```diff
@@ -54,3 +54,5 @@ requires:
     optional: true
   logging:
     interface: loki_push_api
+  nginx-route:
+    interface: nginx-route
```
arturo-seijas: Missing line at EOF

weiiwang01:
> I don't think we should support nginx-route on this charm and on the 12-factor globally
> I'd rather implement as discussed the `ingress-to-nginx-route` or add `ingress` support on nginx-ingress-integrator

A question comes to my mind about using the `traefik-k8s` and `ingress` relation as the application ingress. Let's say we have a charm application called `dqlite` deployed in the `stg-dqlite` model. And we relate the `dqlite` charm with `traefik-k8s` to provide ingress for the `dqlite` charm.

`traefik-k8s` is using the `subdomain` routing mode and we set the `external_hostname` to `canonical.com`, so the traefik hostname for the `dqlite` application is `stg-dqlite-dqlite.canonical.com`. And the plan is to create a DNS CNAME recode from `dqlite.canonical.com` to `stg-dqlite-dqlite.canonical.com`, so users can access the site using the `dqlite.canonical.com`.

But in that case, user's request will carry a `Host: dqlite.canonical.com` header, when arrived at the `traefik-k8s` charm, the host doesn't match `stg-dqlite-dqlite.canonical.com`, so the request will be dropped instead of sending to the `dqlite` charm.

That's why I think even if traefik charm resolved the TLS certificates issue, it still can't provide ingress for our application. I think we should reconsider let our application support `ingress`.

.github/workflows/publish_charm.yaml
weiiwang01: Fixed, thanks!

metadata.yaml
weiiwang01: Added, thanks!

arturo-seijas (approved):
LGTM

gregory-schiano:
> > I don't think we should support nginx-route on this charm and on the 12-factor globally
> > I'd rather implement as discussed the `ingress-to-nginx-route` or add `ingress` support on nginx-ingress-integrator
> 
> A question comes to my mind about using the `traefik-k8s` and `ingress` relation as the application ingress. Let's say we have a charm application called `dqlite` deployed in the `stg-dqlite` model. And we relate the `dqlite` charm with `traefik-k8s` to provide ingress for the `dqlite` charm.
> 
> `traefik-k8s` is using the `subdomain` routing mode and we set the `external_hostname` to `canonical.com`, so the traefik hostname for the `dqlite` application is `stg-dqlite-dqlite.canonical.com`. And the plan is to create a DNS CNAME recode from `dqlite.canonical.com` to `stg-dqlite-dqlite.canonical.com`, so users can access the site using the `dqlite.canonical.com`.
> 
> But in that case, user's request will carry a `Host: dqlite.canonical.com` header, when arrived at the `traefik-k8s` charm, the host doesn't match `stg-dqlite-dqlite.canonical.com`, so the request will be dropped instead of sending to the `dqlite` charm.
> 
> That's why I think even if traefik charm resolved the TLS certificates issue, it still can't provide ingress for our application. I think we should reconsider let our application support `ingress`.

Hence the discussion we got with Observability based on this internally collected feedback (can't share the doc as it's an internal link)
One of the main point was this issue we agreed on:
https://github.com/canonical/charm-relation-interfaces/issues/79

requirements.txt
```diff
@@ -1,3 +1,3 @@
 jsonschema >= 4.17
```
amandahla, line 1: Consider pin jsonschema version since 4.18.0 version might require Rust compiler :|

amandahla (approved):

requirements.txt
weiiwang01: Hi Amanda, as we decided to use the ingress relation instead of nginx-route. I will close this pull request and pin the jsonschema version in the next pull request. Thanks!

PR #33, opened 2023-07-19
Accept proxy environment variables from juju model config

weiiwang01:
Configure proxy-related environmental variables automatically according to the Juju model settings.


src/flask_app.py
```diff
@@ -47,4 +48,9 @@ def flask_environment(self) -> dict[str, str]:
         secret_key_env = f"{FLASK_ENV_CONFIG_PREFIX}SECRET_KEY"
         if secret_key_env not in env:
             env[secret_key_env] = self._charm_state.flask_secret_key
+        for proxy_variable in ("http_proxy", "https_proxy", "no_proxy"):
+            proxy_value = os.environ.get(f"JUJU_CHARM_{proxy_variable.upper()}")
```
gregory-schiano: I'm wondering if the proxy variables shouldn't be part of the charm state ?
@jdkandersson what do you think ?

tests/integration/test_proxy.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+async def test_proxy(build_charm: str, model: Model, pytestconfig: Config, get_unit_ips):
+    """Build and deploy the flask charm."""
+    app_name = "flask-k8s"
+    await model.set_config(
+        {
+            "juju-http-proxy": "http://proxy.test",
+            "juju-https-proxy": "http://proxy.test",
+            "juju-no-proxy": "127.0.0.1,10.0.0.1",
```
gregory-schiano: You should use variables or constants for the proxy parameters to avoid repeating the same string

tests/unit/conftest.py
```diff
@@ -117,3 +118,13 @@ def python_cmd_handler(argv: list[str]) -> ExecResult:
...
+
+
+@pytest.fixture
+def set_environment_variable(request):
+    """Set the environment variables."""
+    environment_variables = request.param
+    os.environ.update(environment_variables)
+    yield
+    for environment_variable in environment_variables:
+        del os.environ[environment_variable]
```
gregory-schiano: Shouldn't you simply keep a copy of the previous env and restore it after yield ?
This will allow overriding existing envrionment variable and restoring them accordingly afterwards instead of removing them

tests/unit/test_flask_app.py
```diff
@@ -43,3 +44,48 @@ def test_flask_env(harness: Harness, flask_config: dict):
...
+            id="https_proxy",
+        ),
+        pytest.param(
+            {"JUJU_CHARM_HTTP_PROXY": "proxy.test", "JUJU_CHARM_HTTPS_PROXY": "proxy.test"},
+            {"http_proxy": "proxy.test", "https_proxy": "proxy.test"},
+            id="http_https_proxy",
+        ),
+    ],
+    indirect=["set_environment_variable"],
+)
```
gregory-schiano: Shouldn't we test the case where we don't set any `JUJU_CHARM_*` proxy variables and we expect no proxy variable in the expected env ?

tests/unit/test_flask_app.py
```diff
@@ -43,3 +44,48 @@ def test_flask_env(harness: Harness, flask_config: dict):
...
+@pytest.mark.usefixtures("set_environment_variable")
+def test_http_proxy(harness: Harness, expected: typing.Dict[str, str]):
+    """
+    arrange: set the juju charm http proxy related environment variables.
```
gregory-schiano: `s/set the//`

tests/unit/test_flask_app.py
```diff
@@ -43,3 +44,48 @@ def test_flask_env(harness: Harness, flask_config: dict):
...
+        pytest.param(
+            {"JUJU_CHARM_HTTP_PROXY": "proxy.test", "JUJU_CHARM_HTTPS_PROXY": "proxy.test"},
+            {"http_proxy": "proxy.test", "https_proxy": "proxy.test"},
+            id="http_https_proxy",
+        ),
+    ],
+    indirect=["set_environment_variable"],
+)
+@pytest.mark.usefixtures("set_environment_variable")
+def test_http_proxy(harness: Harness, expected: typing.Dict[str, str]):
```
gregory-schiano: Do we really need harness to unittest this method ? I don't like the habit of having systematically harness to make unitest as harness acts almost like an integration test

src/flask_app.py
```diff
@@ -47,4 +48,9 @@ def flask_environment(self) -> dict[str, str]:
         secret_key_env = f"{FLASK_ENV_CONFIG_PREFIX}SECRET_KEY"
         if secret_key_env not in env:
             env[secret_key_env] = self._charm_state.flask_secret_key
+        for proxy_variable in ("http_proxy", "https_proxy", "no_proxy"):
```
amandahla, line 50: Would be possible/useful to have the same approach as used by GitHub Runner? Since is a common situation I think would be nice to follow a pattern for that.
https://github.com/canonical/github-runner-operator/blob/main/src/charm.py#L136

gregory-schiano: Indeed that's also a nice approach, which is an alternative to moving these into CharmState as suggested in one of my comments

src/flask_app.py
gregory-schiano: Conversation can be kept in https://github.com/canonical/flask-k8s-operator/pull/33#discussion_r1268096252

src/flask_app.py
jdkandersson: For context, the GitHub runners don't implement the charm state pattern until we refactor it

The approach is ok, it stores the proxies on the charm. It would be better to have them on the state in general since modules might require them and should just have access to them via the state rather than needing something stored on the charm. The proxies are conceptually similar to things like database connection details and so on which should go to the state

With this approach the proxies are not typed either so auto-complete doesn't work and mypy isn't helping

tests/integration/test_cos.py
jdkandersson, line 1: I'm assuming no change here, just moved from another file

tests/integration/test_database.py
jdkandersson, line 1: Also here, assuming just a copy from another module to here

tests/integration/test_proxy.py
jdkandersson: probably would be worth a fixture

tests/integration/test_proxy.py
```diff
@@ -0,0 +1,38 @@
...
+        }
+    )
+    resources = {
+        "flask-app-image": pytestconfig.getoption("--test-flask-image"),
```
jdkandersson: Should this come from a fixture and include an error message if the dev didn't supply it?

tests/integration/test_proxy.py
```diff
@@ -0,0 +1,38 @@
...
+        "statsd-prometheus-exporter-image": "prom/statsd-exporter",
+    }
+    await model.deploy(build_charm, resources=resources, application_name=app_name, series="jammy")
```
jdkandersson, line 29: Shouldn't this be in a fixture?

tests/unit/conftest.py
jdkandersson: I think the [pytest monkeypatch](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.setenv) is designed to handle this and will make sure the environment is restored

src/flask_app.py
weiiwang01: The challenge with storing proxy variables in the state is that they are only used once in the flask-k8s charm, in contrast to the GitHub runner charm, where they are used multiple times.

tests/unit/test_flask_app.py
jdkandersson: It indeed doesn't look like harness is needed

tests/integration/test_proxy.py
weiiwang01: Updated, thanks!

tests/unit/conftest.py
weiiwang01: Updated, but we still need to delete extra environment variables because there isn't a way to replace the entire environment variables.

tests/unit/test_flask_app.py
weiiwang01: That's right! Added, updated, thanks!

tests/unit/test_flask_app.py
weiiwang01: Updated, thanks!

tests/unit/test_flask_app.py
weiiwang01: Removed and using a magic mock instead, thanks!

tests/integration/test_cos.py
weiiwang01: Yep, to increase the parallelization of the integration test.

tests/integration/test_database.py
weiiwang01: Yep, to increase the parallelization of the integration test.

tests/integration/test_proxy.py
weiiwang01: We already have a fixture for normal build and deploy. But for proxy test, we need to set the model configuration prior to the deploy, so we can't reuse that.

tests/unit/conftest.py
jdkandersson: I think monkeypatch has delenv which should restore them after the test

src/flask_app.py
jdkandersson: What is the challenge with that?

tests/integration/test_proxy.py
jdkandersson: ah, perhaps worth a comment?

tests/unit/conftest.py
```diff
@@ -124,7 +124,11 @@ def python_cmd_handler(argv: list[str]) -> ExecResult:
 def set_environment_variable(request):
```
jdkandersson: This should still use pytest monkeypatch

tests/integration/test_proxy.py
weiiwang01: Added, thanks!

tests/integration/test_proxy.py
weiiwang01: Added, thanks!

tests/unit/conftest.py
weiiwang01: That's a good idea, updated, thanks!

tests/unit/conftest.py
weiiwang01: Yes, updated to use the pytest monkeypatch

tests/integration/test_proxy.py
weiiwang01: I don't think we need a fixture since we don't need to resume model configs.

src/flask_app.py
yanksyoon: May I ask why the JUJU_CHARM_ prefix? It seems like the environment variable doesn't have the `JUJU_CHARM` prefix.

weiiwang01: I think this is defined by juju: https://juju.is/docs/sdk/charm-environment-variables

tests/unit/conftest.py
weiiwang01: Incorporating proxies into the charm state requires introducing a new attribute to the interface of the charm state. However, given its limited utility - being used just once - I think it might not worth it.

jdkandersson (dismissed):

src/flask_app.py
weiiwang01: Incorporating proxies into the charm state requires the introduction of a new attribute to the charm state interface. However, given its limited utility - being used just once - I think it might not worth it.

jdkandersson: @gregory-schiano I think we should make a general decision here. I can see the argument - the proxy settings in this case just need to be mapped to a different environment variable.

The general case is that it might be required in a few places, e.g., written to the application config, written into the env etc. In general it seems to best go to the state

amandahla: Just marking @yanksyoon here since same thing is being done in https://github.com/canonical/jenkins-k8s-operator/pull/18

jdkandersson (dismissed):

src/flask_app.py
jdkandersson: @gregory-schiano and I discussed this as a part of another conversation, the proxy should go into the charm state

weiiwang01: Okay, done. The proxy information now exists in the charm state.

jdkandersson (dismissed):

arturo-seijas (dismissed):
LGTM

jdkandersson (approved):

yanksyoon (approved):
LGTM!

PR #37, opened 2023-07-30
Allow flask applications that require database to start

weiiwang01:
The existing design of the database module cannot modify the database configuration unless the Flask application is already initiated. However, for certain Flask applications, initiation is impossible without the proper database configuration in place. This presents a deadlock scenario.

With this pull request, we propose to address this problem by amending the database module to enable it to update the Flask application's database configuration even prior to the application's initiation.


src/charm.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, *args: typing.Any) -> None:
         )
         self._observability = Observability(charm=self, charm_state=self._charm_state)
 
-        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.config_changed, self.on_config_changed)
```
yhaliaw: Why is this changed to a public method?

src/databases.py
```diff
@@ -11,16 +11,13 @@
...
-    FLASK_SERVICE_NAME,
-    FLASK_SUPPORTED_DB_INTERFACES,
-)
-from exceptions import InvalidDatabaseRelationDataError, PebbleNotReadyError
+from constants import FLASK_DATABASE_NAME, FLASK_SUPPORTED_DB_INTERFACES
 
 logger = logging.getLogger(__name__)
 
+if typing.TYPE_CHECKING:
+    from charm import FlaskCharm
```
yhaliaw: Is the if statement needed?
It seems the import is only used for type checking, but in the team's code base we generally do not do this.

arturo-seijas (dismissed):
LGTM

src/charm.py
weiiwang01: Well, because it will be invoked in the `database.py` module, so I have to make it public.

src/databases.py
weiiwang01: This is necessary to avoid circular dependency.

src/charm.py
mthaddon: Could the database module not do a `charm.on.config_changed.emit()` (not sure if that's how it'd reference a charm object, but you get the idea) to avoid this?

mthaddon:
Could you add a description to the PR? It'd be good to have a bit more context on the change.

src/charm.py
mthaddon: Ah, yeah, looks like you should be able to do `self._charm.on.config_changed.emit()`

src/databases.py
mthaddon: I think we should include a comment explaining this, as otherwise it might not be obvious to someone looking at it in the future.

src/charm.py
weiiwang01: That's a great idea, updated, thanks!

weiiwang01:
> Could you add a description to the PR? It'd be good to have a bit more context on the change.

Added, I also merged the proxy pull request to main, so now the change list in the pull request should only include the modification to the database system.

src/databases.py
weiiwang01: Yes, we will have a meeting discussing the circular dependency problem later this day, and hope we can remove them.

src/databases.py
```diff
@@ -137,9 +114,7 @@ def get_uris(self) -> typing.Dict[str, str]:
             # https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/mysql_client/v0/schemas/provider.json
             if not all(data.get(key) for key in ("endpoints", "username", "password")):
                 logger.warning("Incorrect relation data from the data provider: %s", data)
-                raise InvalidDatabaseRelationDataError(
-                    f"Incorrect relation data from the data provider: {data}"
-                )
+                continue
```
amandahla: Why the error is not being raised anymore?
If an application needs database to start and there is no correct relation data, what would happen?

src/databases.py
amandahla: This looks similar (exactly?) to what happened in Synapse where we changed for passing a Pebble instance. I see that you are doing this because the charm needs to call config_change ->  _restart_flask_application. Maybe it's necessary to add the same abstraction here.

jdkandersson:
Just a note - this is still being discuss with @gregory-schiano and myself - the PR shouldn't be merged in its current state

weiiwang01:
> Just a note - this is still being discuss with @gregory-schiano and myself - the PR shouldn't be merged in its current state

Updated with the new design.

src/databases.py
weiiwang01: This has been removed in the latest commit, thanks!

src/databases.py
weiiwang01: In that case, the charm will enter block status with the error message `Webserver configuration check failed, please review your charm configuration or database relation`

src/flask_app.py
```diff
@@ -54,3 +68,53 @@ def flask_environment(self) -> dict[str, str]:
...
+            self._charm.app.status = status
+
+    def restart_flask(self) -> None:
+        """Restart or start the flask service if not started with the latest configuration."""
+        container = self._charm.unit.get_container("flask-app")
+        if not container.can_connect():
+            logger.info("pebble client in the Flask container is not ready")
+            return
+        if not self._charm_state.is_secret_storage_ready:
+            logger.info("secret storage is not initialized, defer config-changed")
```
arturo-seijas: The log msg seems a bit misleading. The event is not deferred

tests/unit/test_charm_state.py
```diff
@@ -44,9 +44,8 @@ def test_charm_state_flask_config(
     harness.begin()
     harness.update_config(charm_config)
     charm_state = CharmState.from_charm(
```
arturo-seijas, line 49: Wouldn't it be better to provide a mocked charm instead of harness for the charm state tests?

src/flask_app.py
weiiwang01: Updated, thanks!

tests/unit/test_charm_state.py
weiiwang01: Yes, it could be, but it will just be too much trouble to mock the configuration retrieving so I opt to use the harness.

arturo-seijas: Doesn't something like the following do the trick?
```
charm = MagicMock(config=charm_config)

Comment (weiiwang01) [reply tests/unit/test_charm_state.py:49]: I haven't thought about that, updated, thanks!

Comment (arturo-seijas) [reply tests/unit/test_charm_state.py:49]: Could you update also the other tests?

File: src/charm.py [hunk 6]
```diff
@@ -29,7 +24,11 @@
 
 
 class FlaskCharm(ops.CharmBase):
-    """Flask Charm service."""
+    """Flask Charm service.
+
+    Args:
```
jdkandersson: Should this be `attrs` instead?

src/charm.py
```diff
@@ -38,22 +37,23 @@ def __init__(self, *args: typing.Any) -> None:
...
             self._update_app_and_unit_status(ops.BlockedStatus(exc.msg))
             return
-        self._flask_app = FlaskApp(charm_state=self._charm_state)
         self._webserver = GunicornWebserver(
             charm_state=self._charm_state,
             flask_container=self.unit.get_container(FLASK_CONTAINER_NAME),
-            flask_app=self._flask_app,
         )
+        self._flask_app = FlaskApp(
+            charm=self, charm_state=self._charm_state, webserver=self._webserver
```
jdkandersson, line 52: Should `FlaskApp` depend on the charm? Also, didn't we discuss that this shouldn't be a class anymore?

src/charm.py
```diff
@@ -38,22 +37,23 @@ def __init__(self, *args: typing.Any) -> None:
...
+        )
+        self._databases = Databases(charm=self, flask_app=self._flask_app)
+        self._charm_state.database_uris = self._databases.get_uris()
```
jdkandersson: This doesn't look right, shouldn't the database_uris be passed as an argument?

tests/unit/test_charm_state.py
weiiwang01: Update other test cases where harness was only used for creating `charm_state`, thanks!

src/charm.py

src/charm.py
weiiwang01: I found out that the `FlaskApp` also need to change the charm status, so just passing the flask container is not sufficient. I think the `FlaskApp` is still a class in the latest design.

src/charm.py
weiiwang01: I can't, another circular dependency.

src/charm.py
jdkandersson: As discussed, the status change should only be done in either the charm class or observers

src/charm.py
```diff
@@ -38,22 +33,22 @@ def __init__(self, *args: typing.Any) -> None:
...
-
         try:
             self._charm_state = CharmState.from_charm(
                 charm=self, secret_storage=self._secret_storage
             )
         except CharmConfigInvalidError as exc:
             self._update_app_and_unit_status(ops.BlockedStatus(exc.msg))
             return
-        self._flask_app = FlaskApp(charm_state=self._charm_state)
         self._webserver = GunicornWebserver(
```
jdkandersson, line 47: This should also not be a class, right?

src/charm.py
jdkandersson: I suggest that we decouple retrieving the database configuration from the database observer so we don't have this issue anymore

jdkandersson: @gregory-schiano

src/charm.py
weiiwang01: Since the `webserver` doesn't cause any trouble, the modification of that class is not in high priority, we can do that later, thanks!

src/charm.py
weiiwang01: FYI, `DataRequires` in `lib/charms/data_platform_libs/v0/data_interfaces.py` can't be instantized twice for the same relation. The singleton property is enforce by `ops` framework.

src/charm.py
weiiwang01: Updated, thanks!

src/charm.py
jdkandersson: What if the database requires instance is passed as an argument to both the database observer and the charm state? That way the state can use it to retrieve the integration data and the observer can use it to listen to events

weiiwang01: I think the code fragmentation problem is worth considering.

jdkandersson: Discussed with Gregory and the `get_uris` can be a function on the database module, just shouldn't be attached to the database observer. There can also be a function to initialise the data structure containing the database requirers classes in the dictionary which can be passed as an argument to the database observer and `get_uris` function. This should address the concern around code fragmentation

weiiwang01: updated.

src/charm.py
```diff
@@ -38,21 +33,26 @@ def __init__(self, *args: typing.Any) -> None:
             args: passthrough to CharmBase.
         """
         super().__init__(*args)
-        self._databases = Databases(charm=self)
         self._secret_storage = SecretStorage(charm=self)
-
+        self._database_requirers = Databases.make_database_requirers(self)
```
jdkandersson: I'm guessing this is a static method, should this just be top level function in the module?

src/charm.py
```diff
@@ -38,21 +33,26 @@ def __init__(self, *args: typing.Any) -> None:
...
         try:
             self._charm_state = CharmState.from_charm(
-                charm=self, secret_storage=self._secret_storage
+                charm=self,
+                secret_storage=self._secret_storage,
+                database_uris=Databases.get_uris(self._database_requirers),
```
jdkandersson: Similar to previous comment, should this just be a top level function in the module?

tests/unit/test_charm.py
```diff
@@ -22,18 +25,24 @@ def test_flask_pebble_layer(harness: Harness) -> None:
     act: start the flask charm and set flask-app container to be ready.
     assert: flask charm should submit the correct flaks pebble layer to pebble.
     """
-    harness.begin_with_initial_hooks()
-    harness.set_can_connect(harness.model.unit.containers["flask-app"], True)
-    harness.framework.reemit()
+    harness.begin()
+    secret_storage = unittest.mock.MagicMock()
+    secret_storage.is_initialized = True
+    secret_storage.get_flask_secret_key.return_value = "0" * 16
```
jdkandersson: Should this be a constant so that it doesn't need to be repeated in the assert?

jdkandersson:
> Just a note - this is still being discuss with @gregory-schiano and myself - the PR shouldn't be merged in its current state

This is getting close now, thanks for the patience and the updates and feedback! We have made important progress with the architecture guidance in the process of reviewing this PR

tests/unit/test_charm.py
weiiwang01: Updated, thanks

src/charm.py
weiiwang01: Updated, thanks

src/charm.py
weiiwang01: Updated, thanks

jdkandersson (dismissed):

src/databases.py
```diff
@@ -32,121 +97,60 @@ class Databases(ops.Object):  # pylint: disable=too-few-public-methods
...
-    def get_uris(self) -> typing.Dict[str, str]:
-        """Compute DatabaseURI and return it.
+        self._charm.unit.status = status
+        if self._charm.unit.is_leader():
+            self._charm.app.status = status
+
+    def _restart_flask(self) -> None:
+        """Restart or start the flask service if not started with the latest configuration."""
+        try:
+            restart_flask(
```
amandahla: Not in this PR but I just saw that restart_flask does not change the unit status if cannot connect to the container or if the secret is not ready. Won't this impact here?

weiiwang01: I think now the `restart_flask` doesn't change the status at all. The status change is now handled by the caller.

jdkandersson (approved):

arturo-seijas (approved):
LGTM

PR #50, opened 2023-09-05
Add database migration script support

weiiwang01:
### Overview

Integrating database migration script support within the flask-k8s charm.

### Rationale

Many Flask applications depend on a database, necessitating the execution of a database migration script prior to the Flask application launch. Previously, users had to manually run the database migration before starting or updating the flask-k8s charm. With this update, users can specify a database migration script within the Flask container via the charm configuration. This allows the charm to automatically manage the database migration process.

### Juju Events Changes

The `update-status` event now re-runs the database migration and launches the Flask application if the migration encountered issues in prior attempts.

### Module Changes

A new module, `database_migration`, has been introduced to manage the database migrations.

### Library Changes

N/A

### Checklist

- [x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The documentation is generated using `src-docs`
- [ ] The documentation for charmhub is updated
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

`src-docs` hasn't been initialized in this repo, and no documents on charmhub in this stage.


config.yaml
```diff
@@ -1,6 +1,18 @@
 # Copyright 2023 Canonical Ltd.
 # See LICENSE file for licensing details.
 options:
+  database_migration_script:
+    type: string
+    description: >-
+      Specifies the path starting from /srv/flask/app that points to a shell script
```
arturo-seijas: s/path/relative path?

test_db_rock/alembic/versions/eca6177bd16a_initial_migration.py
```diff
@@ -0,0 +1,31 @@
+"""Initial migration
```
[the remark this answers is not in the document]
arturo-seijas: You might want to exclude this directory from license checks

src/charm.py
```diff
@@ -169,6 +173,11 @@ def _restart_flask(self) -> None:
         except CharmConfigInvalidError as exc:
             self._update_app_and_unit_status(ops.BlockedStatus(exc.msg))
 
+    def _on_update_status(self, _: ops.HookEvent) -> None:
+        """Handle the update-status event."""
+        if self._database_migration.get_status() == "FAILED":
+            self._restart_flask()
```
arturo-seijas, line 168: Will `_restart_flask` remove the status file?

config.yaml

test_db_rock/alembic/versions/eca6177bd16a_initial_migration.py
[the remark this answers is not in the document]
weiiwang01: I just added the license to those files, since I created them

src/charm.py
weiiwang01: If the database migration run succeeded the status file will be set to "COMPLETED".

arturo-seijas: I meant if there's a `FAILED` status stored. Note that `DatabaseMigration.run` will only execute the migration if the status is `PENDING`

weiiwang01: You're absolutely correct; thank you for pointing that out. I've addressed the issue and implemented additional tests to prevent similar problems.

arturo-seijas (dismissed):
LGTM

config.yaml
```diff
@@ -1,6 +1,18 @@
...
 options:
+  database_migration_script:
+    type: string
+    description: >-
+      Specifies the relative path from /srv/flask/app that points to a shell script
+      executing database migrations for the Flask application. This script is
+      designed to run once for each Flask container unit. However, users must ensure:
+      1. The script can be executed multiple times without issues;
+      2. Concurrent migrations from different units are safe.
+      In case of migration failure, the charm will re-attempt during the
```
amandahla, line 12: I dont know if this was discussed before: In the case of a failure, why not block the charm so the user can review the script?

weiiwang01: It's pretty common for the failure to be caused by temporary network issues, so rerun it can be more user-friendly. I think rerun the script might be more "Kubernetes-ic" since that's how Kubernetes treat errors (`CrashLoopBackOff`)

arturo-seijas (dismissed):

src/charm.py
```diff
@@ -35,6 +36,10 @@ def __init__(self, *args: typing.Any) -> None:
         super().__init__(*args)
         self._secret_storage = SecretStorage(charm=self)
         database_requirers = make_database_requirers(self)
+        self._database_migration = DatabaseMigration(
+            flask_container=self.unit.get_container(FLASK_CONTAINER_NAME),
+            database_migration_script=self.config.get("database_migration_script"),
+        )
```
gregory-schiano: Shouldn't this be done after CharmState is instantiated and use an attribute from the state ?
Pro are:
- If the service need more configuration options in the future we can pass the whole CharmState
- CharmState attributes are validated, which is not necessarily the case here

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+import ops
+from ops.pebble import ExecError
+
+from constants import FLASK_STATE_DIR
+from exceptions import CharmConfigInvalidError
+
+logger = logging.getLogger(__name__)
+
+
+DatabaseMigrationStatus = Literal["PENDING", "COMPLETED", "FAILED"]
```
gregory-schiano: Shouldn't it be an enum ?

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+        if (
+            self.get_completed_script() is not None
+            and self.script is not None
+            and self.script != self.get_completed_script()
+        ):
+            raise CharmConfigInvalidError(
+                f"database migration script {self.get_completed_script()!r} "
+                f"has been executed successfully in the current flask container,"
+                f"updating database-migration-script in config has no effect"
+            )
```
gregory-schiano: That's a bit strange to have a method that only raise exception, couldn't it return a boolean and manage what to do from the caller side

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+            CharmConfigInvalidError: if the database migration run failed.
+        """
+        if self.get_status() in (self.PENDING, self.FAILED) and self.script:
+            logger.info("execute database migration script: %s", repr(self.script))
+            try:
+                self._container.exec(
+                    ["/bin/bash", "-xeo", "pipefail", self.script],
+                    environment=environment,
+                    working_dir="/srv/flask/app",
+                ).wait()
```
gregory-schiano: Shouldn't we put the charm in Maintenance status with an explicit message while it's running ?

src/webserver.py
```diff
@@ -141,9 +149,11 @@ def update_config(self, flask_environment: dict[str, str], is_webserver_running:
                 "Webserver configuration check failed, "
                 "please review your charm configuration or database relation"
             ) from exc
+        self._database_migration.run(flask_environment)
```
gregory-schiano: I'm wondering if that should be done there @jdkandersson  what do you think ?
I think that either:
- We move the run on the FlaskApp after "update_config" is called
- Or we rename this method as it now does more than just update the config

tests/unit/test_database_migration.py
```diff
@@ -0,0 +1,98 @@
...
+
+from charm_state import CharmState
+from constants import FLASK_CONTAINER_NAME
+from database_migration import DatabaseMigration
+from exceptions import CharmConfigInvalidError
+from flask_app import FlaskApp
+from webserver import GunicornWebserver
+
+
+def test_database_migration(harness: Harness):
```
gregory-schiano, line 18: Shouldn't this test be even more scoped ? It's calling `restart_flask` where I would expect unittests that covers:
`get_status`, `set_status`, `get_completed_script`, `set_completed_script`, `run` and `check_completed`

Then the webserver test modified to asserts that the corresponding methods are called with the right arguments

src/charm.py
weiiwang01: I wonder if this attribute is only used once in the `DatabaseMigration`, should this be put in the charm state. But anyway I will put this in the charm state just for now. Thanks!

src/database_migration.py
weiiwang01: I find more convenient to provide them as constant like `DatabaseMigration.PENDING`, `DatabaseMigration.COMPLETED` and `DatabaseMigration.FAILED`. And probably as safe.

src/database_migration.py
weiiwang01: Yes, you are right. Updated, thanks!

src/webserver.py
weiiwang01: Yes, I think we should move the database migration to `FlaskApp`. The only drawback of that approach is that the database migration will run after the server is reloaded via signal. But I assume if the server need to be reloaded via signal, it means the server has already started, which probably mean that the database migration is already completed.

tests/unit/test_database_migration.py
weiiwang01: Okay, I will add more tests that target different method of `DatabaseMigration`. This test is more for the overall behavior of database migration.

src/charm.py
jdkandersson: +1 to using CharmState

src/charm.py
```diff
@@ -169,6 +173,11 @@ def _restart_flask(self) -> None:
         except CharmConfigInvalidError as exc:
             self._update_app_and_unit_status(ops.BlockedStatus(exc.msg))
 
+    def _on_update_status(self, _: ops.HookEvent) -> None:
+        """Handle the update-status event."""
+        if self._database_migration.get_status() == self._database_migration.FAILED:
```
jdkandersson: Should `self._database_migration.FAILED` just be a constant rather than a property of the class

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+        FAILED: one of the possible database migration status.
+        PENDING: one of the possible database migration status.
+    """
+
+    _STATUS_FILE = FLASK_STATE_DIR / "database-migration-status"
+    _COMPLETED_SCRIPT_FILE = FLASK_STATE_DIR / "completed-database-migration"
+
+    COMPLETED: DatabaseMigrationStatus = "COMPLETED"
+    FAILED: DatabaseMigrationStatus = "FAILED"
+    PENDING: DatabaseMigrationStatus = "PENDING"
```
jdkandersson: As noted elsewhere, these would be better to be associated with the module rather than the class

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def __init__(self, flask_container: ops.Container, database_migration_script: str | None):
```
jdkandersson: Under what conditions would `database_migration_script` be `None` please? I guess if the user doesn't have a migration script defined?

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+        return None
+
+    def set_completed_script(self, script_path: str) -> None:
+        """Set the database migration script that has completed in the current container.
+
+        Args:
+            script_path: The completed database migration script in the current container.
+        """
+        self._container.make_dir(self._COMPLETED_SCRIPT_FILE.parent, make_parents=True)
+        self._container.push(self._COMPLETED_SCRIPT_FILE, script_path)
```
jdkandersson: What is the reason for needing these please?

jdkandersson: perhaps they can be private?

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+        """Get the database migration script."""
+        return self._script
+
+    def get_status(self) -> DatabaseMigrationStatus:
+        """Get the database migration run status.
+
+        Returns:
+            One of "PENDING", "COMPLETED", or "FAILED".
+        """
+        return cast(
```
jdkandersson: Would it be clearer to split this up into a few lines to first handle the file doesn't exist case and then read?

src/database_migration.py
```diff
@@ -0,0 +1,142 @@
...
+from constants import FLASK_STATE_DIR
+from exceptions import CharmConfigInvalidError
+
+logger = logging.getLogger(__name__)
+
+
+DatabaseMigrationStatus = Literal["PENDING", "COMPLETED", "FAILED"]
+
+
+class DatabaseMigration:
```
jdkandersson, line 33: Please review the methods on this class for what should be public versus private

src/webserver.py
jdkandersson: Yeh this doesn't seem to be the right place to do it.

I'm also wondering whether it is a flask or webserver responsibility for running the migration? It seems more a charm responsibility to execute the migration script and then start the webserver/ flask since they are essentially distinct responsibilities

src/charm.py
weiiwang01: I found that put the constants in the class namespace is easier to use since users don't need to import addition variables, and the constants is always available where the `DatabaseMigration` is used.

src/database_migration.py
weiiwang01: This is to show user a warning message when they change the database migration script after the execution is completed inside the container. Because in that case, the charm will not run the new database migration script, and it's important to let user know their changes on the `database_migration_script` have no effect.

src/database_migration.py
weiiwang01: This function is only used once, so I have inlined this function, thanks!

src/charm.py

src/database_migration.py
weiiwang01: See: https://github.com/canonical/flask-k8s-operator/pull/50#discussion_r1318248094, thanks!

src/database_migration.py
weiiwang01: Yes, if user doesn't config the `database_migration_script` charm configuration.

src/database_migration.py
jdkandersson: I was more talking about whether `set_completed_script` and `get_completed_script` could be private methods?

src/database_migration.py
weiiwang01: Yes, I think the setters can be private in this case, updated.

src/database_migration.py
weiiwang01: See https://github.com/canonical/flask-k8s-operator/pull/50#discussion_r1320814651, thanks!

tests/integration/test_db_migration.py
```diff
@@ -0,0 +1,48 @@
...
+
+async def test_db_migration(
+    flask_db_app: Application,
+    model: juju.model.Model,
+    get_unit_ips,
+):
+    """
+    arrange: build and deploy the flask charm.
+    act: deploy the database and relate it to the charm.
+    assert: requesting the charm should return a correct response indicate
```
jdkandersson, line 23: Is there anything more authoritative we can query? E.g., a table existing that should have been created in the migration? Just checking that the app is running could be met by just not running the migration script, right?

weiiwang01: If the database migration script doesn't create the table, the request will result in an error. Check the implementation of the testing flask application https://github.com/canonical/flask-k8s-operator/blob/81352e4659e1f4df56b1378c5723da3bd04ba315/test_db_rock/app.py

jdkandersson: so if the flask application implementation changes then the test doesn't fail thats why it would be better to capture in the test that everything worked as expected. E.g., consider the case where someone just changes the test application - it isn't clear that this would actually constitute a change in what the test checks for

src/webserver.py
jdkandersson: Discussed with @gregory-schiano, because the script gets the flask environment it seems to be appropriate to run it from the flask app class

src/database_migration.py
jdkandersson: Also discussed with @gregory-schiano, because our tooling already enforces descriptions on class attributes and because enums are a general concept, we will propose a standard for the team to use enums instead of literals in these cases so we have a consistent approach across the team

src/charm.py
```diff
@@ -7,12 +7,13 @@
 import logging
 import typing
 
-import ops
+import ops.testing
```
amandahla: What is using ops.testing?

src/charm.py
```diff
@@ -44,12 +46,20 @@ def __init__(self, *args: typing.Any) -> None:
         except CharmConfigInvalidError as exc:
             self._update_app_and_unit_status(ops.BlockedStatus(exc.msg))
             return
-        self._webserver = GunicornWebserver(
+
+        self._database_migration = DatabaseMigration(
+            flask_container=self.unit.get_container(FLASK_CONTAINER_NAME),
```
amandahla, line 50: I think I didn't find where is checked if container can_connect. Isn't that needed?

src/charm_state.py
```diff
@@ -213,9 +219,17 @@ def from_charm(
             )
             error_field_str = " ".join(f"flask_{f}" for f in error_fields)
             raise CharmConfigInvalidError(f"invalid configuration: {error_field_str}") from exc
+        database_migration_script = charm.config.get("database_migration_script")
```
amandahla, line 222: Couldn't this be checked on database_migration since you have a container instance that can be used to check the existence of the file? The script is supposed to be there, right?

src/constants.py
```diff
@@ -3,8 +3,13 @@
...
 
+import pathlib
+
 FLASK_CONTAINER_NAME = "flask-app"
 FLASK_SERVICE_NAME = "flask-app"
 FLASK_ENV_CONFIG_PREFIX = "FLASK_"
 FLASK_DATABASE_NAME = "flask-app"
 FLASK_SUPPORTED_DB_INTERFACES = {"mysql_client": "mysql", "postgresql_client": "postgresql"}
+FLASK_BASE_DIR = pathlib.Path("/srv/flask")
+FLASK_APP_DIR = pathlib.Path("/srv/flask/app")
```
amandahla: Minor/nitpick: FLASK_BASE_DIR usage instead of "/srv/flask" ?

src/database_migration.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+
+    def _set_status(self, status: DatabaseMigrationStatus) -> None:
+        """Set the database migration run status.
+
+        Args:
+            status: One of "PENDING", "COMPLETED", or "FAILED".
+        """
+        self._container.make_dir(self._STATUS_FILE.parent, make_parents=True)
+        self._container.push(self._STATUS_FILE, source=status)
```
amandahla: container.push has a make_dirs=True parameter. Would that work here instead of make_dir?

src/database_migration.py
```diff
@@ -0,0 +1,125 @@
...
+        Raises:
+            CharmConfigInvalidError: if the database migration run failed.
+        """
+        if self.get_status() in (self.PENDING, self.FAILED) and self.script:
+            logger.info("execute database migration script: %s", repr(self.script))
+            try:
+                self._container.exec(
+                    ["/bin/bash", "-xeo", "pipefail", self.script],
+                    environment=environment,
+                    working_dir="/srv/flask/app",
```
amandahla: Minor/nitpick: Isn't there a constant for /srv/flask/app? the FLASK_APP_DIR.

tests/unit/test_database_migration.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    arrange: none
```
amandahla, line 20: Please, could you give more details about this test? 
I see that you set Flaskapp and then restart Flask three times always checking the database_migration_history but I think I didn't follow why there is no "database-migration-2.sh" in the last check.

amandahla:
Really nice PR :rocket: . Left some comments.

src/charm.py
weiiwang01: I must accidentally imported that, removed, thanks!

src/charm.py
weiiwang01: Yes, it is needed, but it has already been checked in the `restart_flask` function of `flask_app.py`.

src/charm_state.py
weiiwang01: This is not to check if the file exist, this is to avoid user provide configuration like `../../foobar` to escape from the base directory.

src/constants.py
weiiwang01: Updated, thanks!

src/database_migration.py
weiiwang01: Okay, updated

src/database_migration.py
weiiwang01: That's really convenient! Updated, thanks!

src/database_migration.py
weiiwang01: Yep, I forgot that, updated, thanks!

tests/integration/test_db_migration.py
weiiwang01: Okay, I added an endpoint to inspect the tables in database.

tests/unit/test_database_migration.py
weiiwang01: The rule is that when the database migration completed, it will be an error if user attempts to update the `database_migration_script` configuration and the updated value will be ignored. The last `database-migration-2.sh` is to simulate that.

jdkandersson (dismissed):

src/charm_state.py
```diff
@@ -213,9 +219,17 @@ def from_charm(
...
+        if database_migration_script:
+            database_migration_script = os.path.normpath(FLASK_APP_DIR / database_migration_script)
+            if not database_migration_script.startswith(str(FLASK_APP_DIR)):
+                raise CharmConfigInvalidError(
+                    "database_migration_script is not inside /srv/flask/app"
```
yhaliaw: Should the `/srv/flask/app` be hardcoded?

weiiwang01: Yes, indeed, updated, thanks!

jdkandersson (dismissed):

amandahla (dismissed):
LGTM

src/database_migration.py
```diff
@@ -0,0 +1,129 @@
...
+logger = logging.getLogger(__name__)
+
+
+class DatabaseMigrationStatus(str, enum.Enum):
+    """Database migration status.
+
+    Attrs:
+        COMPLETED: one of the possible database migration status.
+        FAILED: one of the possible database migration status.
+        PENDING: one of the possible database migration status.
```
yanksyoon:
```suggestion
        COMPLETED: A status denoting a successful database migration.
        FAILED: A status denoting an unsuccessful database migration.
        PENDING: A status denoting a pending database migration..
```
A bit of a trivial docstring, I think it can be good as is :D

src/database_migration.py
```diff
@@ -0,0 +1,129 @@
...
+        Args:
+            environment: Environment variables that's required for the run.
+
+        Raises:
+            CharmConfigInvalidError: if the database migration run failed.
+        """
+        if (
+            self.get_status() in (DatabaseMigrationStatus.PENDING, DatabaseMigrationStatus.FAILED)
+            and self.script
+        ):
```
yanksyoon:
```suggestion
        if (
            self.get_status() not in (DatabaseMigrationStatus.PENDING, DatabaseMigrationStatus.FAILED)
            or not self.script
        ):
        return
```
A small nitpicky suggestion to reduce indentation :D

yanksyoon (dismissed):
LGTM! Minor suggestions that can be done in next PR at your preference!

jdkandersson (dismissed):

src/database_migration.py
weiiwang01: Updated, thanks!

src/database_migration.py
weiiwang01: Yes, updated, thanks!

jdkandersson (approved):

yanksyoon (approved):
LGTM, thanks for the updates!

PR #56, opened 2023-09-25
Change the base directory from /srv/flask to /flask

weiiwang01:
### Overview

Modify the base directory from `/srv/flask` to `/flask` to simplify the OCI image structure. Additionally, force the WSGI path to `app:app` to align with the `flask-framework` Rockcraft extension.

### Rationale

This update ensures better alignment between the `flask-k8s` charm and the `flask-framework` Rockcraft extension.

### Juju Events Changes

None.

### Module Changes

None.

### Library Changes

None.

### Checklist

- [x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [x] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [x] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [x] The documentation is generated using `src-docs`
- [x] The documentation for charmhub is updated
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


arturo-seijas (dismissed):
LGTM

test_rock/app.py
```diff
@@ -99,12 +99,3 @@ def postgresql_status():
 def get_env():
     """Return environment variables"""
     return jsonify(dict(os.environ))
-
-
-app2 = Flask("app2")
```
jdkandersson, line 104: It looks like some tests are being removed? Is it because we have simplified the design and hence made them redundant?

src/constants.py
```diff
@@ -10,6 +10,6 @@
 FLASK_ENV_CONFIG_PREFIX = "FLASK_"
 FLASK_DATABASE_NAME = "flask-app"
 FLASK_SUPPORTED_DB_INTERFACES = {"mysql_client": "mysql", "postgresql_client": "postgresql"}
-FLASK_BASE_DIR = pathlib.Path("/srv/flask")
+FLASK_BASE_DIR = pathlib.Path("/flask")
```
amandahla, line 13: Minor/nitpicking:
Constants were much discussed in this PR:
https://github.com/canonical/synapse-operator/pull/36

Should we keep using this file/adding constants here or should FLASK_BASE_DIR be created on webserver.py and then imported on the unit test?

test_rock/app.py
weiiwang01: Yes, I have removed the `webserver_wsgi_path` setting and it's now always `app:app` to be aligned with the extension.

jdkandersson (approved):

arturo-seijas (approved):

src/constants.py
weiiwang01: I think the `FLASK_*_DIR` is not just used in the `webserver.py`, it's used in the `charm_state.py` and `database_migration.py` as well. That's why I think it can be put in the `constants.py`.

